PR #1, opened 2019-09-10
Pass resource limits to JM/TM containers through Downward API

functicons:
The limits will be used by the containers to set the default Java heap
size for JM/TM processes.


config/samples/flinkoperator_v1alpha1_flinkjobcluster.yaml
```diff
@@ -18,7 +18,7 @@ metadata:
   name: flinkjobcluster-sample
 spec:
   image:
-    name: flink:1.8.1
+    name: gcr.io/google.com/hadoop-cloud-dev/flink:1.8.1-dataproc
```
aniket486: Remove

PR #3, opened 2019-09-15
Add build dependency check in Makefile

functicons:


scripts/check_build_deps.sh
```diff
@@ -0,0 +1,49 @@
...
+
+echo "Checking build dependencies..."
+
+# Go
+go_version=$(go version | go version | grep -P -o '\d\.\d+\.\d+')
+if [[ ! "${go_version}" > "1.12." ]]; then
+  echo "Error: Go 1.12+ is required for the build."
+  echo "Please install it by following the instructions at ${kustomize_url}"
+  exit 1
+fi
```
medb: IS it possible to extract this into a function?

PR #21, opened 2019-09-25
Fix Go version check problem

functicons:


scripts/check_build_deps.sh
```diff
@@ -8,7 +8,7 @@ echo "Checking build dependencies..."
 # Go
 go_url="https://golang.org/doc/install"
 go_version=$(go version | grep -P -o '\d\.\d+')
-if [[ ! "${go_version}" > "1.12" ]]; then
+if [[ ! "${go_version}" == 1.12* && ! "${go_version}" > "1.12" ]]; then
```
medb: It doesn't look that string comparison `"${go_version}" > "1.12"` will do the right thing here. You need to split version on components and compare them as numbers.

PR #32, opened 2019-09-28
Check job pod status in addition to job status to update CR status

### LINKED ISSUE #30: Job status is marked as Running in CR when the job pod is still Pending

functicons:
Job status could be marked as Running when the pod is still pending for
being scheduled, this caused we mark FlinkCluster status as Running when
it is still creating.

Fix #30

### DIFF
--- a/controllers/flinkcluster_observer.go
+++ b/controllers/flinkcluster_observer.go
@@ -78,10 +78,9 @@ func (observer *_ClusterStateObserver) observe(
 		if client.IgnoreNotFound(err) != nil {
 			log.Error(err, "Failed to get the cluster resource")
 			return err
-		} else {
-			log.Info("Observed cluster", "cluster", "nil")
-			observedCluster = nil
 		}
+		log.Info("Observed cluster", "cluster", "nil")
+		observedCluster = nil
 	} else {
 		log.Info("Observed cluster", "cluster", *observedCluster)
 		observedState.cluster = observedCluster
@@ -94,10 +93,9 @@ func (observer *_ClusterStateObserver) observe(
 		if client.IgnoreNotFound(err) != nil {
 			log.Error(err, "Failed to get JobManager deployment")
 			return err
-		} else {
-			log.Info("Observed JobManager deployment", "state", "nil")
-			observedJmDeployment = nil
 		}
+		log.Info("Observed JobManager deployment", "state", "nil")
+		observedJmDeployment = nil
 	} else {
 		log.Info("Observed JobManager deployment", "state", *observedJmDeployment)
 		observedState.jmDeployment = observedJmDeployment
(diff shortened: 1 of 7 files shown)


controllers/flinkcluster_observer.go
```diff
@@ -135,35 +136,78 @@ func (observer *_ClusterStateObserver) observe(
...
+		if client.IgnoreNotFound(err) != nil {
+			log.Error(err, "Failed to get job pods")
+			return err
+		} else {
+			log.Info("Observed job pods", "pods", "nil")
+			observedJobPods = nil
+		}
+	} else {
+		log.Info("Observed job pods", "pods", observedJobPods.Items)
+		if len(observedJobPods.Items) == 1 {
```
hongyegong: Should we log something here if no job pod is found at all?

controllers/flinkcluster_observer.go
```diff
@@ -272,3 +316,15 @@ func (observer *_ClusterStateObserver) observeJob(
...
+func (observer *_ClusterStateObserver) observeJobPods(observedJobPod *corev1.PodList) error {
+	var clusterName = observer.request.Name
+	var jobName = getJobName(observer.request.Name)
+	var inNamespace = client.InNamespace(observer.request.Namespace)
+	var matchingLabels client.MatchingLabels = map[string]string{
+		"app":      "flink",
+		"cluster":  clusterName,
+		"job-name": jobName,
+	}
+	return observer.k8sClient.List(context.Background(), observedJobPod, inNamespace, matchingLabels)
```
hongyegong: Can we make this a variable just so that it's more readable?

controllers/flinkcluster_reconciler.go
```diff
@@ -213,11 +214,23 @@ func (reconciler *_ClusterReconciler) reconcileJob() error {
 	var log = reconciler.log
 	var desiredJob = reconciler.desiredState.job
 	var observedJob = reconciler.observedState.job
-
-	if desiredJob != nil && observedJob == nil {
-		return reconciler.createJob(desiredJob)
+	var observedClusterComponents = reconciler.observedState.cluster.Status.Components
+	if desiredJob != nil {
+		if observedJob == nil {
```
hongyegong: how's this possible under the if condition above?

controllers/flinkcluster_updater.go
```diff
@@ -74,10 +74,16 @@ func (updater *_ClusterStatusUpdater) updateClusterStatusIfChanged() error {
...
 func (updater *_ClusterStatusUpdater) deriveClusterStatus() flinkoperatorv1alpha1.FlinkClusterStatus {
 	var status = flinkoperatorv1alpha1.FlinkClusterStatus{}
-	var totalComponents = 3
-	var readyComponents = 0
+	var runningComponents = 0
 	var recordedClusterStatus = &updater.observedState.cluster.Status
 
+	var totalComponents = 0
+	if updater.observedState.cluster.Spec.JobSpec != nil {
+		totalComponents = 4
```
hongyegong: Can we make this hardcoded numbers some static variables with some meaningful names?

controllers/flinkcluster_observer.go
functicons: The previous line will log empty list if not job pod is found, and it is a normal condition, not an error.

controllers/flinkcluster_reconciler.go
functicons: This is a normal case, "desiredJob != nil" means we want to create a job, while "observedJob == nil" means we haven't created yet, then the action will be just to create one.

controllers/flinkcluster_updater.go
functicons: I feel it is better to add comments.

controllers/flinkcluster_observer.go
functicons: Done

PR #33, opened 2019-10-03
Re-queue the reconcile after reconciliation failed.

hongyegong:
Reconciliation event should be immediately rescheduled if a controller enters an error condition and returns an error in the reconcile method.

### DIFF
--- a/controllers/flinkcluster_controller.go
+++ b/controllers/flinkcluster_controller.go
@@ -156,7 +156,7 @@ func (handler *_FlinkClusterHandler) Reconcile(
 	if err != nil {
 		log.Error(err, "Failed to reconcile")
 		return ctrl.Result{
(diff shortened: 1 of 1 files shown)


controllers/flinkcluster_controller.go
```diff
@@ -154,7 +155,10 @@ func (handler *_FlinkClusterHandler) Reconcile(
 	err = reconciler.reconcile()
 	if err != nil {
 		log.Error(err, "Failed to reconcile")
-		return ctrl.Result{}, err
+		return ctrl.Result{
+                	RequeueAfter: time.Second,
```
functicons: 1s is too short, maybe 5s is better.

hongyegong: Done.

PR #36, opened 2019-10-05
Add status update events for FlinkCluster CR

### LINKED ISSUE #11: Add events to the FlinkCluster CR

functicons:
Fix #11

### DIFF
--- a/controllers/flinkcluster_updater.go
+++ b/controllers/flinkcluster_updater.go
@@ -27,6 +27,7 @@ import (
 	"github.com/go-logr/logr"
 	flinkoperatorv1alpha1 "github.com/googlecloudplatform/flink-operator/api/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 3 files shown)


controllers/flinkcluster_updater.go
```diff
@@ -47,31 +49,98 @@ func (updater *_ClusterStatusUpdater) updateClusterStatusIfChanged() error {
...
+	if oldStatus.Components.TaskManagerDeployment.State !=
+		newStatus.Components.TaskManagerDeployment.State {
+		updater.createStatusChangeEvent(
+			"TaskManager deployment",
+			oldStatus.Components.TaskManagerDeployment.State,
+			newStatus.Components.TaskManagerDeployment.State)
+	}
+
+	// Job.
+	if oldStatus.Components.Job == nil && newStatus.Components.Job != nil {
```
hongyegong: is there a case where old is not nil but new is nil?

functicons: No, there shouldn't be such a case.

PR #45, opened 2019-10-10
Add ingress resource

elanv:
It would be nice to provide limited access with http(s) via ingress for dashboard access or for REST API connections.

Ingress is enabled with operator flag --ingress-host-format

ex) flink-operator --ingress-host-format {{$clusterName}}.example.com


functicons:
Thanks for the contribution! Will review later today.

functicons:
A high-level question about the `ingress-host-format`, why is it a command line flag? I feel it makes deployment more complex. Can it be part of the CRD?

elanv:
Thank you for quick response!

Users who want the ingress feature will probably have their own domain, so it may be convenient to preconfigure the domain format and let the operator automatically create an ingress host in that format. So, in this PR, the cluster administrator provides the hostname format as the operator setting to automatically create an ingreass host. Since I thought it was an operator-level setup, I received the host format as a command line flag rather than a CRD.

I think the way to get the operator setting can be implemented in other ways such as configMap instead of the command line flag. As such an operator configuration system is not yet implemented, I have passed it as a command line flag. Later, when the configuration system is developed, it is likely that we can add an implementation to get the configuration in that way.

api/v1alpha1/flinkcluster_default.go
```diff
@@ -26,6 +26,7 @@ func _SetDefault(cluster *FlinkCluster) {
 	_SetJobManagerDefault(&cluster.Spec.JobManagerSpec)
 	_SetTaskManagerDefault(&cluster.Spec.TaskManagerSpec)
 	_SetJobDefault(cluster.Spec.JobSpec)
+	_SetJobManagerIngressDefault(cluster.Spec.JobManagerSpec.Ingress)
 }
```
functicons: This should be part of _SetJobManagerDefault.

controllers/flinkcluster_converter.go
```diff
@@ -213,6 +218,76 @@ func getDesiredJobManagerService(
 	return jobManagerService
 }
 
+// Gets the desired JobManager service spec from a cluster spec.
```
functicons: nit: s/service/ingress

controllers/flinkcluster_controller.go
```diff
@@ -123,6 +133,15 @@ func (handler *_FlinkClusterHandler) Reconcile(
 	} else {
 		log.Info("Desired state", "JobManager service", "nil")
 	}
+	if desiredState.jmIngress != nil {
+		log.Info("Desired state", "JobManager ingress", *desiredState.jmIngress)
+	} else {
+		var msg string
+		if observedState.cluster != nil && observedState.cluster.Spec.JobManagerSpec.Ingress != nil && config.IngressHostFormat == "" {
```
functicons: This logic should be moved to the validating webhook, reject the create request if the format is missing. Since we haven't implemented the validating webhook, it is okay to put it here for now, but add a TODO comment here please.

functicons:
I just merged #50, could you sync? thanks!

elanv:
Thank you for your review! Sorry for the late response. 
I understand your concern and I'll reflect your opinion.

api/v1alpha1/flinkcluster_types.go
```diff
@@ -110,6 +110,21 @@ type JobManagerPorts struct {
 	UI *int32 `json:"ui,omitempty"`
 }
 
+// JobManagerIngress defines ingress of JobManager
+type JobManagerIngressSpec struct {
+	// Ingress host format. ex) {{$clusterName}}.example.com
+	IngressHostFormat string `json:"ingressHostFormat"`
```
functicons: "Ingress" is redundant in this context, just use "HostFormat".

controllers/flinkcluster_converter_test.go
```diff
@@ -274,6 +285,57 @@ func TestGetDesiredClusterState(t *testing.T) {
...
+						}},
+				},
+			}},
+			TLS: []extensionsv1beta1.IngressTLS{{
+				Hosts: []string{"flinkjobcluster-sample.example.com"},
+			}},
+		},
+	}
+
+	assert.Assert(t, desiredState.JmDeployment != nil)
```
functicons: JmIngress

api/v1alpha1/flinkcluster_types.go
elanv: Fixed.

controllers/flinkcluster_converter_test.go
elanv: Fixed.

functicons:
The code LGTM. Did you test it? in which environment? I'll test it in GKE before merging.

elanv:
Thanks!

I have tested in Azure AKS, my test cluster.

elanv:
Sorry, there was a mistake in the cluster role resource name and it is fixed.

api/v1alpha1/flinkcluster_types.go
```diff
@@ -110,6 +110,21 @@ type JobManagerPorts struct {
 	UI *int32 `json:"ui,omitempty"`
 }
 
+// JobManagerIngress defines ingress of JobManager
+type JobManagerIngressSpec struct {
+	// Ingress host format. ex) {{$clusterName}}.example.com
+	HostFormat string `json:"hostFormat"`
+
```
functicons: Can we make HostFormat optional? This will allow accessing the ingress with external IP address without domain name.

api/v1alpha1/flinkcluster_types.go
```diff
@@ -282,6 +303,18 @@ type JobStatus struct {
...
 
+// JobManagerIngressStatus defines the status of a JobManager ingress.
+type JobManagerIngressStatus struct {
+	// The name of the Kubernetes ingress resource.
+	Name string `json:"name"`
+
+	// The state of the component.
+	State string `json:"state"`
+
+	// The URL of ingress.
```
functicons: nit: s/URL/URLs

controllers/flinkcluster_updater.go
```diff
@@ -205,6 +219,47 @@ func (updater *_ClusterStatusUpdater) deriveClusterStatus() flinkoperatorv1alpha
...
+					if host != "" {
+						urls = append(urls, "https://"+host)
+					}
+				}
+			}
+		} else {
+			for _, rule := range observedJmIngress.Spec.Rules {
+				if rule.Host != "" {
+					urls = append(urls, "http://"+rule.Host)
+				}
```
functicons: If we make host optional, the IP address will be dynamically allocated, so here we need to use IP https://godoc.org/k8s.io/api/core/v1#LoadBalancerIngress

api/v1alpha1/flinkcluster_types.go
elanv: I agree.

controllers/flinkcluster_updater.go
```diff
@@ -239,6 +245,22 @@ func (updater *_ClusterStatusUpdater) deriveClusterStatus() flinkoperatorv1alpha
...
+		// If ingress does not have host, get ip from status
+		if len(urls) == 0 && observedJmIngress.Status.LoadBalancer.Ingress != nil {
+			var scheme string
+			if useTLS {
+				scheme = "https://"
+			} else {
+				scheme = "http://"
+			}
+			for _, ingress := range observedJmIngress.Status.LoadBalancer.Ingress {
+				if ingress.IP != "" {
```
functicons: Should we add host as well if it is present?

elanv: Yes, all of the observed information looks good to include.

controllers/flinkcluster_updater.go
```diff
@@ -205,6 +219,72 @@ func (updater *_ClusterStatusUpdater) deriveClusterStatus() flinkoperatorv1alpha
...
+
+		if len(observedJmIngress.Spec.TLS) > 0 {
+			useTLS = true
+		}
+
+		if useTLS {
+			for _, tls := range observedJmIngress.Spec.TLS {
+				for _, host := range tls.Hosts {
+					if host != "" {
+						urls = append(urls, "https://"+host)
```
functicons: IIUC, urls represents the actual status, why do we include the host from the spec here? because below we use `len(urls) > 0` to determine the state of the Ingress, this makes it always Ready when host is provided, unable to detect failure. I think we should only include URLs from observedJmIngress.Status.LoadBalancer.Ingress.

elanv: Thank you for your good feedback.

Firstly, the hostname of the ingress status is the hostname of the load balancer itself, so it will always provide the hostname of the LB separately from the host of the spec. Since the user provides the host in the ingress spec to access the host defined differently for each ingress resource, not the hostname of the LB, the URL intended by the user should be the host of the spec.

Secondly, as you say, the ready state is right by judging the status of an ingress resource. This is because the LB IP or hostname in ingress status is updated from blank to the actual LB value when the LB providing the ingress is determined. (And either IP or hostname is being used. Link: https://github.com/kubernetes/ingress-nginx/blob/04e2ad8fcd51b0741263a37b8e7424ca3979137c/internal/ingress/status/status.go#L218)

I initially tried to check the status as the ingress resource creation is complete because the ingress resource status may not be updated and may remain blank depending on the environment or method of installing the ingress controller. However, this is rare in a cloud environment, so in your opinion it seems reasonable to change to ready state when the LB value of status is filled. In order to properly check the status of ingress, it seems correct to configure the cluster so that the ingress status is updated on the cluster administrator side.

So what if we change two things?
1. The jobmanager ingress status is determined to be ready when the LB value of the ingress resource status is updated.
2. Specify the LB IP or hostname of ingress resource status in jobmanager ingress status urls only when jobmanager ingress hostFormat is not provided.

Note) There are some cases where the status of an ingress resource remains blank.
1.When installing with the commonly used helm chart stable / nginx-ingress and installing with default value without providing controller.publishService.enabled option (https://github.com/helm/charts/blob/master/ stable / nginx-ingress / values.yaml # L79)
2.If your ingress controller is configured to use the host network (https://github.com/kubernetes/ingress-nginx/blob/master/docs/deploy/baremetal.md)

functicons: Thanks for your detailed explanation. The solution sounds good to me. I just tested it in Minikube, it worked. I'll test it in GKE tomorrow.

functicons:
Could you update the CRD doc with ingress? Thanks! https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/docs/crd.md

elanv:
I have added JobManager ingress explanation to CRD doc.

functicons:
On a second thought, could you remove ingress from sample YAML files? because it might not work out of the box in some environments. We want to keep samples simple and always successful.

elanv:
> On a second thought, could you remove ingress from sample YAML files? because it might not work out of the box in some environments. We want to keep samples simple and always successful.

I agree. I removed it from samples.

functicons:
Could you resolve conflicts before I merge?

elanv:
I have merged. There seems to be no conflict.

PR #50, opened 2019-10-11
Add unit test for computing desired state from CR

functicons:


functicons:
> Can we put the unit test under a testing folder?

I'd like to, but this is not the convention of Go.

PR #53, opened 2019-10-13
Submit job after Flink API is ready

### LINKED ISSUE #42: Getting Flink job ID always fails when job is running and succeeds when job is finished
Noticed a strange behavior in the log, in the controller for job clusters, getting Flink job ID always fails when job is running and succeeds when job is finished.

We can do a test with session cluster, launch a client container manually calls the jobs endpoint of the Flink REST API.

functicons:
Fix #42

### DIFF
--- a/controllers/flinkcluster_observer.go
+++ b/controllers/flinkcluster_observer.go
@@ -18,14 +18,12 @@ package controllers
 
 import (
 	"context"
-	"encoding/json"
+	"errors"
 	"fmt"
-	"io/ioutil"
-	"net/http"
-	"time"
 
 	"github.com/go-logr/logr"
 	flinkoperatorv1alpha1 "github.com/googlecloudplatform/flink-operator/api/v1alpha1"
+	"github.com/googlecloudplatform/flink-operator/flinkclient"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


controllers/flinkcluster_reconciler.go
```diff
@@ -210,28 +208,33 @@ func (reconciler *_ClusterReconciler) deleteService(
...
-			var jmServiceReady = observedClusterComponents.JobManagerService.State ==
-				flinkoperatorv1alpha1.ClusterComponentState.Ready
-			var tmDeploymentReady = observedClusterComponents.TaskManagerDeployment.State ==
-				flinkoperatorv1alpha1.ClusterComponentState.Ready
-			if jmDeploymentReady && jmServiceReady && tmDeploymentReady {
-				return reconciler.createJob(desiredJob)
+			// If the observed Flink job status list is not nil (e.g., emtpy list), it means
+			// Flink REST API server is up and running. It is the source of truth of whether
+			// we can submit a job.
+			if reconciler.observedState.flinkJobStatusList != nil {
```
hongyegong: just observed? here and everywhere in this method.

controllers/flinkcluster_observer.go
```diff
@@ -163,90 +165,77 @@ func (observer *_ClusterStateObserver) observeJob(
...
 
-// Gets Flink job ID through Flink REST API.
-func (observer *_ClusterStateObserver) getFlinkJobID(url string) *string {
+// Observes Flink jobs through Flink API (instead of Kubernetes jobs through
+// Kubernetes API).
+//
+// This needs to be done after the cluster is running and before the job is
+// submitted, because we use it to detect whether the Flink API server is up
+// and running.
+func (observer *_ClusterStateObserver) observeFlinkJobs(
```
hongyegong: Not sure if observe[foo] is a converntion is kebernetes API, but this method does bunch of things like updating job lists and job ID, the method name itself doesn't appear to match with what method itself is doing.
Maybe consider breaking method down further or replace with a more specific method name?

functicons: "updating job lists and job ID" is applied to an intermediate object, which is used to just hold the observed state, not applied to the cluster CR.

controllers/flinkcluster_reconciler.go
functicons: I don't understand your question. If the observed flinkJobStatusList is not nil, it means we have call Flink API successfully, then we can submit job to Flink.

Issue #62, opened 2019-10-17
Delay deleting job cluster resources after the job finishes


functicons:
Currently when the job of a job cluster finishes or fails, the controller immediately starts deleting the JM/TM resources, we want to delay the process for a while (e.g., 5 mins by default) to allow checking logs and metrics. Maybe make the delay configurable as a property of the CR.

PR #63, opened 2019-10-18
Delay deleting cluster resources by 5 mins after the job finishes

functicons:
After the job finishes, the cluster will transit into Stopping state
and keep in that state for 5 mins, then start deleting cluster
resources.


functicons:
Fix #62

Issue #58, opened 2019-10-16
Support config map


functicons:
https://ci.apache.org/projects/flink/flink-docs-stable/ops/deployment/kubernetes.html#session-cluster-resource-definitions

PR #76, opened 2019-10-25
Add configMap for cluster configuration

elanv:
Related issue #58

* Changes
  * Add a configMap resource and mount it to use flink-conf.yaml and the log configuration file
  * Removes existing environment variables that duplicate functions

* Improvement needs
  * docker-entrypoint replacement


functicons:
Thanks for the contribution! Some high-level questions first:

1) IIUC, if flinkProperties is provided in the CR, it will create a ConfigMap and override the default Flink config file, which means if a property is in the default config but not in the flinkProperties, it will be lost. In the existing env variable based solution, it doesn't override the default config, it appends the custom properties at the end of the config. I think this is a change of behavior and we probably want to keep the default config.

2) What do you mean by "Improvement needs: docker-entrypoint replacement"?

controllers/flinkcluster_converter.go
```diff
@@ -406,6 +364,58 @@ func getDesiredTaskManagerDeployment(
...
+	var clusterNamespace = flinkCluster.ObjectMeta.Namespace
+	var clusterName = flinkCluster.ObjectMeta.Name
+	var flinkProperties = flinkCluster.Spec.FlinkProperties
+	var jmPorts = flinkCluster.Spec.JobManager.Ports
+	var configMapName = getConfigMapName(clusterName)
+	var labels = map[string]string{
+		"cluster": clusterName,
+		"app":     "flink",
+	}
+	var flinkProps = map[string]string{
```
elanv: Connection configurations should reflect the actual deployment.

elanv:
1. I reflected the behavior of the official docker image in the implementation of this PR. As far as I can see, most of the settings in flink-conf.yaml in the official docker image are commented out, and only the ports and addresses that are required for flink JM and TM connections are actually valid.(https://github.com/apache/flink/blob/release-1.8.1/flink-dist/src/main/resources/flink-conf.yaml)
In docker-entrypoint, flink-conf.yaml may be added if there is no port or address set. This PR is also implemented to add the actual deployed address and port to filnk-conf.yaml so that there is no problem with the connection between JM and TM even if the user does not provide it. flinkProperties is used as an addition to those preferences. If the user sets the address and port, the code needs to be changed so that the connection information is not overridden so that the value is different from the actual deployment environment. (https://github.com/elanv/flink-on-k8s-operator/blob/feature/configmap/controllers/flinkcluster_converter.go#L389-L397)

2. docker-entrypoint.sh of the official image mainly consists of a script to set the address and port related to flink connection, so when using configMap, there seems to be a lot of unnecessary content. And I have the opinion that we need to improve the pod spec's commands and args in case of using custom image as well as official image.

controllers/flinkcluster_converter.go
```diff
@@ -330,43 +312,19 @@ func getDesiredTaskManagerDeployment(
...
-				},
-			},
-		},
-		{
-			Name:  "FLINK_PROPERTIES",
-			Value: getFlinkProperties(flinkCluster.Spec.FlinkProperties),
-		},
-	}
-	envVars = append(envVars, flinkCluster.Spec.EnvVars...)
+	// Make Volume, VolumeMount to use configMap data for flink-conf.yaml, if flinkProperties is provided.
```
functicons: Here it says "if flinkProperties is provided", but it seems not true, configMap is always created, is my understanding correct?

controllers/flinkcluster_converter.go
```diff
@@ -330,43 +312,19 @@ func getDesiredTaskManagerDeployment(
...
-			ValueFrom: &corev1.EnvVarSource{
-				ResourceFieldRef: &corev1.ResourceFieldSelector{
-					ContainerName: "taskmanager",
-					Resource:      "limits.cpu",
-					Divisor:       resource.MustParse("1m"),
-				},
-			},
-		},
-		{
-			Name: "TASK_MANAGER_MEMORY_LIMIT",
```
functicons: Previously, I was using downward API to get CPU and memory limit, then in my custom image, I used the env variables to update the heapSize and numOfSlots in flink config. I plan to make a upstream PR, to allow passing the values to the official image. So, here can we keep the TASK_MANAGER_CPU_LIMIT and TASK_MANAGER_MEMORY_LIMIT?

controllers/flinkcluster_converter.go
```diff
@@ -406,6 +364,58 @@ func getDesiredTaskManagerDeployment(
 	return taskManagerDeployment
 }
 
+// Gets the desired TaskManager deployment spec from a cluster spec.
+func getDesiredConfigMap(
```
functicons: Since JM/TM deployment depend on the configmap, do we need to wait configmap to be ready first, then create deployments?

controllers/flinkcluster_converter.go
```diff
@@ -541,3 +558,67 @@ func isStopDelayExpired(
...
+	}
+	confMount = &corev1.VolumeMount{
+		Name:      flinkConfigMapVolume,
+		MountPath: flinkConfigMapPath,
+	}
+	return confVol, confMount
+}
+
+// TODO: Wouldn't it be better to create a file, put it in an operator image, and read from them?.
+// Provide logging profiles
```
functicons: I think it is a good idea to have default config files included in the operator image.

controllers/flinkcluster_converter.go
elanv: The configmap is always created. I'll edit the comment.

controllers/flinkcluster_converter.go
elanv: Yes, I will preserve them.

controllers/flinkcluster_converter.go
elanv: The configMap will be created almost at the same time as the request, but there will be a slight delay since the pod's attempt to mount the configMap requires some sort of pod scheduling first. So is it not sufficient to request the configMap creation before JM, TM rather than controlling the order in the operator? Even if the configMap creation is delayed, to my knowledge, if the pod attempts to mount the configMap and fails because the configMap does not exist, it will repeat the mount retry. As far as I tested, the pod did not fail or restart in this process and kept its initial state "ContainerCreating".

controllers/flinkcluster_observer.go
```diff
@@ -135,6 +136,21 @@ func (observer *ClusterStateObserver) observe(
 		observed.tmDeployment = observedTmDeployment
 	}
 
+	// (Optional) configMap.
```
functicons: Remove "(Optional)"

controllers/flinkcluster_reconciler.go
```diff
@@ -74,6 +74,11 @@ func (reconciler *ClusterReconciler) reconcile() (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 
+	err = reconciler.reconcileConfigMap()
```
functicons: Can we move reoncileConfigMap() to the beginning so that it will be created before JM/TM deployments.

functicons:
There is a unit test failure, could you fix it?

```
--- FAIL: TestGetDesiredClusterState (0.00s)
    flinkcluster_converter_test.go:553: assertion failed: 
        --- desiredState.ConfigMap
        +++ expectedConfigMap
          v1.ConfigMap{
          	TypeMeta:   v1.TypeMeta{},
          	ObjectMeta: v1.ObjectMeta{Name: "flinkjobcluster-sample-configmap", Namespace: "default", Labels: map[string]string{"app": "flink", "cluster": "flinkjobcluster-sample"}, OwnerReferences: []v1.OwnerReference{{APIVersion: "flinkoperator.k8s.io/v1alpha1", Kind: "FlinkCluster", Name: "flinkjobcluster-sample", Controller: &true, BlockOwnerDeletion: &false}}},
          	Data: map[string]string{
          		"flink-conf.yaml": strings.Join({
          			... // 2 identical lines
          			"jobmanager.rpc.port: 6123",
          			"query.server.port: 6125",
        - 			"rest.port: 8081",
          			"taskmanager.numberOfTaskSlots: 1",
          			"",
          		}, "\n"),
          		"log4j-console.properties": "log4j.rootLogger=INFO, console\nlog4j.logger.akka=INFO\nlog4j.logger.org.apache.kafka=INFO\nlog4j.logger.org.apache.hadoop=INFO\nlog4j.logger.org.apache.zookeeper=INFO\nlog4j.appender.console=org.apache.log4j.ConsoleAppender\nlog4j.appender.console.layout=org.apache.log4j.PatternLayout\nlog4j.appender.console.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss,SSS} %-5p %-60c %x - %m%n\nlog4j.logger.org.apache.flink.shaded.akka.org.jboss.netty.channel.DefaultChannelPipeline=ERROR, console",
          		"logback-console.xml":      "<configuration>\n    <appender name=\"console\" class=\"ch.qos.logback.core.ConsoleAppender\">\n        <encoder>\n            <pattern>%d{yyyy-MM-dd HH:mm:ss.SSS} [%thread] %-5level %logger{60} %X{sourceThread} - %msg%n</pattern>\n        </encoder>\n    </appender>\n    <root level=\"INFO\">\n        <appender-ref ref=\"console\"/>\n    </root>\n    <logger name=\"akka\" level=\"INFO\">\n        <appender-ref ref=\"console\"/>\n    </logger>\n    <logger name=\"org.apache.kafka\" level=\"INFO\">\n        <appender-ref ref=\"console\"/>\n    </logger>\n    <logger name=\"org.apache.hadoop\" level=\"INFO\">\n        <appender-ref ref=\"console\"/>\n    </logger>\n    <logger name=\"org.apache.zookeeper\" level=\"INFO\">\n        <appender-ref ref=\"console\"/>\n    </logger>\n    <logger name=\"org.apache.flink.shaded.akka.org.jboss.netty.channel.DefaultChannelPipeline\" level=\"ERROR\">\n        <appender-ref ref=\"console\"/>\n    </logger>\n</configuration>",
          	},
          	BinaryData: nil,
          }
```

Issue #79, opened 2019-10-29
Support cancel job with savepoint


functicons:
We want to support cancel job with auto savepoint. The approach I am thinking about is to add a property `cancelled` in jobSpec and support updating CR with the property. After detecting the `cancelled` set to true, the controller calls Flink API to take savepoint, then cancel the job. We might introduce another property to control whether to force cancel the job if taking savepoint failed.

PR #82, opened 2019-10-30
Support cancel job with savepoint

functicons:
Add a property `cancelRequested` in jobSpec and allow it to be updated.


api/v1alpha1/flinkcluster_validate.go
```diff
@@ -55,14 +55,45 @@ func (v *Validator) ValidateCreate(cluster *FlinkCluster) error {
...
+func (v *Validator) checkCancelRequested(
+	old *FlinkCluster, new *FlinkCluster) (bool, error) {
+	if old.Spec.Job == nil || new.Spec.Job == nil {
+		return false, nil
+	}
+	var restartJob = (old.Spec.Job.CancelRequested != nil && *old.Spec.Job.CancelRequested) &&
+		(new.Spec.Job.CancelRequested == nil || !*new.Spec.Job.CancelRequested)
+	if restartJob {
+		return false, fmt.Errorf(
+			"updating cancelRequested from true to false is not allowed")
```
hongyegong: Why cant this be just restart is not allowed after cancle?

controllers/flinkcluster_reconciler.go
```diff
@@ -428,23 +424,50 @@ func (reconciler *ClusterReconciler) getFlinkJobID() string {
...
 	}
+	if !reconciler.isJobFinished() {
+		return reconciler.stopJob(jobID)
+	}
+	log.Info("Job has finished, no need to cancel", "jobID", jobID)
+	return nil
+}
+
+// Stops a job through Flink REST API.
+func (reconciler *ClusterReconciler) stopJob(jobID string) error {
```
hongyegong: Worth creating a wrapper method here? seems like it's only called once in this method and whatever pre-logic here can just go there.

controllers/flinkcluster_updater.go
```diff
@@ -371,13 +374,19 @@ func (updater *ClusterStatusUpdater) deriveClusterStatus(
...
 			status.Components.Job.State = v1alpha1.JobState.Succeeded
 			jobFinished = true
 			jobSucceeded = true
 		} else {
 			status.Components.Job = nil
 		}
+	} else if recordedJobStatus != nil && len(recordedJobStatus.Name) > 0 {
+		status.Components.Job = recordedJobStatus.DeepCopy()
+		status.Components.Job.State = v1alpha1.JobState.Cancelled
+		jobFinished = true
```
hongyegong: Jobfinished doesn't seem to be homogeneous with jobcancled.

api/v1alpha1/flinkcluster_validate.go
functicons: We'll support restart after finishing #84.

functicons:
Fix #79

PR #112, opened 2019-11-15
Support GCP service account as a first class property

### LINKED ISSUE #110: Support Flink cluster-wide GCP service account
I'm thinking about adding a field in the FlinkCluster CR, adding a reference to the Secret which holds the GCP service account key file, then automatically mount it into all containers including JM/TM and job client. This will make it much easier to configure service account.

functicons:
Users need to create a Secret with the service account key file, then add the gcpConfig property in the FlinkCluster CR, the controller will automatically create volume and mount it into all JM/TM/Job
containers, it also set GOOGLE_APPLICATION_CREDENTIALS to the service account key file path.

Fix #110


controllers/flinkcluster_util.go
```diff
@@ -61,6 +61,11 @@ func getJobName(clusterName string) string {
 	return clusterName + "-job"
 }
 
+// Gets GCP service account volume name
+func getGCPServiceAccountVolumeName(clusterName string) string {
+	return clusterName + "-gcp-sa"
```
hongyegong: "gcp-sa-volume"?

config/samples/flinkoperator_v1alpha1_flinkjobcluster.yaml
```diff
@@ -37,10 +37,5 @@ spec:
     className: org.apache.flink.streaming.examples.wordcount.WordCount
     args: ["--input", "./README.txt"]
     parallelism: 2
-    savepointsDir: /tmp
```
hongyegong: Are these related to this change?

controllers/flinkcluster_converter.go
```diff
@@ -758,6 +786,30 @@ func getFlinkConfRsc(clusterName string) (*corev1.Volume, *corev1.VolumeMount) {
 	return confVol, confMount
 }
 
+func convertGCPConfig(gcpConfig *v1alpha1.GCPConfig) (*corev1.Volume, *corev1.VolumeMount) {
+	if gcpConfig == nil {
+		return nil, nil
+	}
+
+	// Service account.
```
hongyegong: This comment doesn't explain anything.

controllers/flinkcluster_converter.go
```diff
@@ -389,6 +401,11 @@ func getDesiredTaskManagerDeployment(
 		PeriodSeconds:       60,
 		FailureThreshold:    5,
 	}
+	var saVolume, saMount = convertGCPConfig(clusterSpec.GCPConfig)
+	if saVolume != nil && saMount != nil {
+		volumes = append(volumes, *saVolume)
+		volumeMounts = append(volumeMounts, *saMount)
+	}
```
hongyegong: take out the podspec as you did for jobmanager and job to keep code style consistency?

controllers/flinkcluster_util.go
functicons: Actually, this function is not used, just removed it.

config/samples/flinkoperator_v1alpha1_flinkjobcluster.yaml
functicons: Not really, just want to keep the sample simple.

controllers/flinkcluster_converter.go
functicons: Done.

PR #140, opened 2019-12-20
Add Slack channel in README

functicons:


burkaygur:
@functicons, how can I get an invite to the Slack channel?

PR #154, opened 2020-01-15
Update CRD doc 

elanv:
Added afterJobCancelled to cleanupPolicy and fixed typos.


functicons:
Thanks! Could you also update the [alpha doc](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/docs/crd_v1alpha1.md)?

PR #167, opened 2020-02-06
Remove cert mgr

hongyegong:


Makefile
```diff
@@ -80,17 +80,19 @@ install: manifests
 	kubectl apply -f config/crd/bases
 
 # Deploy cert-manager which is required by webhooks of the operator.
-cert-manager:
-	bash scripts/deploy_cert_manager.sh
+cert-generator:
+	bash scripts/cert_generator.sh --service flink-operator-webhook-service --secret webhook-server-cert -n $(FLINK_OPERATOR_NAMESPACE)
```
functicons: Rename it to generate_cert.sh

scripts/cert_generator.sh
```diff
@@ -0,0 +1,131 @@
...
+            ;;
+    esac
+    shift
+done
+
+[ -z ${service} ] && service="flink-operator-webhook-service"
+[ -z ${secret} ] && secret="webhook-server-cert"
+[ -z ${namespace} ] && namespace="flink-operator-system"
+
+# Create the namespace to store cert in
```
functicons: We also need to check if the ${secret} already exits.

scripts/cert_generator.sh
```diff
@@ -0,0 +1,131 @@
...
+            shift
+            ;;
+        *)
+            usage
+            ;;
+    esac
+    shift
+done
+
+[ -z ${service} ] && service="flink-operator-webhook-service"
```
functicons: Always use "[[ ]]" instead of "[ ]".

functicons: Because you `set -u` for this script, if `service` is not provided, `[ -z ${service} ]` would return error because you are trying to evaluate a undefined variable.

functicons: Because this script is general, not specific to Flink operator, it is not appropriate to set default value like "flink-operator-webhook-service".

scripts/cert_generator.sh
```diff
@@ -0,0 +1,131 @@
...
+usage() {
+    cat <<EOF
+Generate certificate suitable for admission webhook service.
+This script uses k8s' CertificateSigningRequest API to a generate a
+certificate signed by k8s CA suitable for admission webhook services.
+This requires permissions to create and approve CSR. See
+https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster for
+detailed explantion and additional instructions.
+The server key/cert k8s CA cert are stored in a k8s secret.
+usage: ${0} [OPTIONS]
```
functicons: s/usage/Usage. Add a newline before this line.

scripts/cert_generator.sh
```diff
@@ -0,0 +1,131 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+usage() {
+    cat <<EOF
+Generate certificate suitable for admission webhook service.
```
functicons: Remove "suitable".

scripts/cert_generator.sh
```diff
@@ -0,0 +1,131 @@
...
+This script uses k8s' CertificateSigningRequest API to a generate a
+certificate signed by k8s CA suitable for admission webhook services.
```
functicons: ditto.

scripts/cert_generator.sh
```diff
@@ -0,0 +1,131 @@
...
+    exit 1
+fi
+echo ${serverCert} | openssl base64 -d -A -out ${tmpdir}/server-cert.crt
+
+
+# create the secret with CA cert and server cert/key
+kubectl create secret generic ${secret} \
+        --from-file=tls.key=${tmpdir}/server-key.pem \
+        --from-file=tls.crt=${tmpdir}/server-cert.crt \
+        --dry-run -o yaml |
```
functicons: Replace these 2 lines as:

```
        --dry-run -o yaml \
    | kubectl -n ${namespace} apply -f -
```

scripts/cert_generator.sh
```diff
@@ -0,0 +1,131 @@
...
+      --namespace        Namespace where webhook service and secret reside.
+      --secret           Secret name for CA certificate and server certificate/key pair.
+EOF
+    exit 1
+}
+
+while [[ $# -gt 0 ]]; do
+    case ${1} in
+        --service)
+            service="flink-operator-webhook-service"
```
functicons: Why hard code to "flink-operator-webhook-service"?

Makefile
```diff
@@ -80,17 +80,19 @@ install: manifests
 	kubectl apply -f config/crd/bases
 
 # Deploy cert-manager which is required by webhooks of the operator.
-cert-manager:
-	bash scripts/deploy_cert_manager.sh
+cert-generator:
```
functicons: Rename it to webhook-cert.

Makefile
functicons: This should be updated to generate_cert.sh.

scripts/generate_cert.sh
```diff
@@ -0,0 +1,146 @@
...
+if [[ ! -x "$(command -v openssl)" ]]; then
+  echo "openssl not found"
+  exit 1
+fi
+
+csrName=${service}.${namespace}
+tmpdir=$(mktemp -d)
+echo "creating certs in tmpdir ${tmpdir} "
+
+cat << EOF >> ${tmpdir}/csr.conf
```
functicons: Why is this append instead of overwrite?

scripts/generate_cert.sh
```diff
@@ -0,0 +1,146 @@
...
+EOF
+
+openssl genrsa -out ${tmpdir}/server-key.pem 2048
+openssl req -new -key ${tmpdir}/server-key.pem -subj "/CN=${service}.${namespace}.svc" -out ${tmpdir}/server.csr -config ${tmpdir}/csr.conf
+
+# clean-up any previously created CSR for our service. Ignore errors if not present.
+kubectl delete csr ${csrName} 2> /dev/null || true
+
+# create  server cert/key CSR and send to k8s API
+cat << EOF | kubectl create -f -
```
functicons: What if the ${csrName} already exits?

scripts/generate_cert.sh
```diff
@@ -0,0 +1,146 @@
...
+  exit 1
+fi
+echo ${serverCert} | openssl base64 -d -A -out ${tmpdir}/server-cert.crt
+
+# create the secret with CA cert and server cert/key
+kubectl create secret generic ${secret} \
+  --from-file=tls.key=${tmpdir}/server-key.pem \
+  --from-file=tls.crt=${tmpdir}/server-cert.crt \
+  --dry-run -o yaml |
+  kubectl -n ${namespace} apply -f -
```
functicons: Move "|" from the previous line to the beginning of this line. Make it clearer, this is a pipe.

scripts/generate_cert.sh
functicons: Have you tested? It will fail I think, you need to use `kubectl apply -f` in this case.

scripts/generate_cert.sh
```diff
@@ -0,0 +1,146 @@
...
+  echo "ERROR: After approving csr ${csrName}, the signed certificate did not appear on the resource. Giving up after 10 attempts." >&2
+  exit 1
+fi
+echo ${serverCert} | openssl base64 -d -A -out ${tmpdir}/server-cert.crt
+
+# create the secret with CA cert and server cert/key
+kubectl create secret generic ${secret} \
+  --from-file=tls.key=${tmpdir}/server-key.pem \
+  --from-file=tls.crt=${tmpdir}/server-cert.crt \
+  --dry-run -o yaml \
```
functicons: This looks weird to me, why use dry run to generate the yaml then apply it? why don't create it directly?

scripts/generate_cert.sh
hongyegong: Are you talking about this line? -cat << EOF | kubectl create -f -
It works as expected for me. Since there is a pipe there i don't see why it should not work.

scripts/generate_cert.sh
hongyegong: This logic is like - create if doesnt exist, then just apply the change without raising any error. Create itself raises error when secret  already exists.

PR #196, opened 2020-03-13
Separate CRD uninstall from helm uninstall

elanv:
Separated CRD uninstall and operator undeploy.
When the CRD is deleted, all deployed FlinkClusters are deleted, but IMO, it may not be the intended result in some situations.

For the helm chart, I suggest applying the helm3 crd installation method instead of the template method to prevent CRD deletion when chart uninstallation.
(note: https://helm.sh/docs/topics/chart_best_practices/custom_resource_definitions/)


functicons:
Because users still need to uninstall CRD, I'm thinking about this structure:

```
deploy-crd:
    ...
deploy-controller:
   ...
deploy: deploy-crd deploy-controller:
   ...
undeploy-crd:
   ...
undeploy-controller:
   ...
undeploy:
   ...
```

In this way, we can keep `deploy` and `undeploy` semantics unchanged.

elanv:
Okay, I agree to your opinion.

What do you think about helm chart? In my opinion it would be better to recommend delete CRD manually and provide a guide how to do. There may be an option to provide CRD deletion as an option when uninstall a chart release, but I am not sure how to implement.

functicons:
Sounds good to me regarding helm chart.

elanv:
I have changed commit for only helm chart.

Issue #194, opened 2020-03-13
In the  submit job script , when the job 'CANCELED', the job pod is running, and crd will not change the status.


Mrart:
When I cancelled the job in flink ui. the job pod log 
`Waiting for job to finish...
Waiting for response...
No running jobs.
No scheduled jobs.
---------------------- Terminated Jobs -----------------------
12.03.2020 03:46:38 : 27e55a40581c628b77741549762b6600 : Flink add sink (CANCELED)
--------------------------------------------------------------`
but job pod still running ,  and crds will not get the flinkclusters status, and do some thing change. In the per-job mode, it seems need stop the job pod? 
As the code need change like this:
`function wait_for_job() {
	while true; do
		echo -e "\nWaiting for job to finish..."
		list_jobs
		if list_jobs | grep -e "(FINISHED)" -e "(FAILED)" -e "(CANCELED)"; then
			echo -e "\nJob has terminated, exiting..."
			break
		fi
		sleep 30
	done
}`
@functicons

functicons:
Right, we need to add CANCELED. Thanks for this feedback. Could you make the change? I'll review it.

Mrart:
Ok ， I will do this. I  will test more.

Mrart:
As I test, When I add the 'cancelled' status in the script. And when I cancelled the job ,the cluster will deleted all workload. and the cr status is "stoped", It is intented ? Or I need do other things?

PR #198, opened 2020-03-16
Add CANCELED status when wait_for_job

Mrart:
Add CANCELED status when wait_for_job #194

### DIFF
--- a/controllers/flinkcluster_submit_job_script.go
+++ b/controllers/flinkcluster_submit_job_script.go
@@ -66,7 +66,7 @@ function wait_for_job() {
 	while true; do
 		echo -e "\nWaiting for job to finish..."
 		list_jobs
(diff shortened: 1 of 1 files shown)


controllers/flinkcluster_submit_job_script.go
```diff
@@ -66,7 +66,7 @@ function wait_for_job() {
 	while true; do
 		echo -e "\nWaiting for job to finish..."
 		list_jobs
-		if list_jobs | grep -e "(FINISHED)" -e "(FAILED)"; then
+		if list_jobs | grep -e "(FINISHED)" -e "(FAILED)" -e "(CANCELLED)"; then
```
functicons: I tested, seems it should be `CANCELED` instead of `CANCELLED`.

Mrart: yes, It's my fault

PR #200, opened 2020-03-18
Support multiple operator deployment

elanv:
Currently, Makefile supports only one operator deployment on one k8s cluster. It seems that the following template generation feature is required to deploy multiple operators.

- Avoid duplicate cluster-scoped resource names
- Support selector to prevent admission webhook call interference (namespaceSelector is used in this PR)

To do this, in this PR build-overlay action is added to the Makefile and webhook patch is added to kustomize config.

Operators can be deployed like:
```bash
make deploy  WATCH_NAMESPACE=dev RESOURCE_PREFIX=dev-
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,10 @@ IMG ?= gcr.io/flink-operator/flink-operator:latest
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 # The Kubernetes namespace in which the operator will be deployed.
 FLINK_OPERATOR_NAMESPACE ?= flink-operator-system
+# Prefix for Kubernetes resource names. When deploying multiple operators, make sure that the names of cluster-scoped resources are not duplicated.
+RESOURCE_PREFIX ?= flink-operator-
+# The Kubernetes namespace to limit watching.
+WATCH_NAMESPACE ?=
 
 #################### Local build and test ####################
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -81,29 +85,51 @@ install: manifests
 
 # Deploy cert-manager which is required by webhooks of the operator.
 webhook-cert:
-	bash scripts/generate_cert.sh --service flink-operator-webhook-service --secret webhook-server-cert -n $(FLINK_OPERATOR_NAMESPACE)
+	bash scripts/generate_cert.sh --service $(RESOURCE_PREFIX)webhook-service --secret webhook-server-cert -n $(FLINK_OPERATOR_NAMESPACE)
 
 config/default/manager_image_patch.yaml:
 	cp config/default/manager_image_patch.template config/default/manager_image_patch.yaml
 
+build-overlay:
```
functicons: Add a comment for this target.

Makefile
```diff
@@ -81,29 +85,51 @@ install: manifests
...
+ifneq ($(WATCH_NAMESPACE),)
+	cd config/deploy \
+			&& sed -E -i.bak  "s/(\-\-watch\-namespace\=)/\1$(WATCH_NAMESPACE)/" manager_auth_proxy_patch.yaml \
+			&& kustomize edit add patch webhook_namespace_selector_patch.yaml \
+			|| true
+endif
+	sed -E -i.bak "s/resources:/bases:/" config/deploy/kustomization.yaml
+	rm config/deploy/*.bak
+
+template: build-overlay
```
functicons: ditto.

Makefile
```diff
@@ -81,29 +85,51 @@ install: manifests
...
-			| sed -e "s/flink-operator-system/$(FLINK_OPERATOR_NAMESPACE)/g" \
-			| sed -e "s/--watch-namespace=/--watch-namespace=$(WATCH_NAMESPACE)/" \
+undeploy-controller: build-overlay
+	kubectl kustomize config/deploy \
+			| sed -e "s/$(RESOURCE_PREFIX)system/$(FLINK_OPERATOR_NAMESPACE)/g" \
 			| kubectl delete -f - \
 			|| true
+ifneq ($(WATCH_NAMESPACE),)
+	kubectl label ns $(WATCH_NAMESPACE) flink-operator-namespace-
+endif
```
functicons: I don't quite understand this command.

Makefile
elanv: Done.

Makefile
elanv: Done.

Makefile
elanv: I have commented on the reason. This part removes the label set of the watch-target namespace for namespaceSelector when operator deployed.

PR #224, opened 2020-04-14
Using Semantic Versioning for Chart version.

kinderyj:
I got an exception when trying to push this charts to the my server.
It show me that the chart version shoud use the Semantic Versioning. If the version is not "Semantic Versioning"， the chart server will not recognize it.
Semantic version Guide: https://semver.org/


kinderyj:
> Have you tested your change? This is a problem that has been realized, but I think just changing this won't be enough for it to work.

Yes, I have tested this change. With this change, I could push the chart to my chart server, and then the server will make it as a valid chart, and that , I can download and install  the operator from my chart server.

Issue #231, opened 2020-04-17
Job restart fails 


elanv:
Even if restartPolicy is configured, failed job is not be recovered. It seems wait_for_job() of job submitter causes the result. Operator recreates the job resource, but the job status become failed by wait_for_job because cancelled or failed jobs are listed, then operator cancels the running job to restart it again. Restart and fail are repeated by operator. 

I have reproduced this by cancelling job in Fink UI, but if the job fails for some reason, the results would be probably the same.

CR yaml: 
```
  job:
    autoSavepointSeconds: 60
    restartPolicy: FromSavepointOnFailure
    cleanupPolicy:
      afterJobFails: KeepCluster
```

flinkCluster status (repeated restart and fail):
```
Status:
  Components:
    Job:
      Name:                       t2-job
      Restart Count:              4
      Savepoint Generation:       2
      Savepoint Location:         ...
      State:                      Running
Events:
  Type    Reason        Age                From           Message
  ----    ------        ----               ----           -------
  Normal  StatusUpdate  18m (x5 over 20m)  FlinkOperator  (combined from similar events): Job status changed: Failed -> Running
  Normal  StatusUpdate  18m (x2 over 18m)  FlinkOperator  Job status changed: Running -> Failed
  Normal  StatusUpdate  18m (x2 over 18m)  FlinkOperator  Job status changed: Failed -> Running
```

Job pod logs:
```

Job has been submitted with JobID 3005678f5f70c30a16cf7e7a7aee61f5

Waiting for job to finish...
Waiting for response...
------------------ Running/Restarting Jobs -------------------
17.04.2020 16:47:02 : 3005678f5f70c30a16cf7e7a7aee61f5 : State machine job (RUNNING)
--------------------------------------------------------------
No scheduled jobs.
---------------------- Terminated Jobs -----------------------
17.04.2020 16:44:10 : 310e00855925191c794af47e8d78ac48 : State machine job (CANCELED)
17.04.2020 16:45:53 : 6e6d5fddff883d4a1a5a842821240106 : State machine job (CANCELED)
17.04.2020 16:46:18 : 7505841fd9e62c3278088e3db960153e : State machine job (CANCELED)
17.04.2020 16:46:40 : 4ea984043dd609f05b4c99020f2993b8 : State machine job (CANCELED)
--------------------------------------------------------------
17.04.2020 16:44:10 : 310e00855925191c794af47e8d78ac48 : State machine job (CANCELED)
[...]

PR #232, opened 2020-04-20
Check active job first to decide job state

elanv:
It seems we need to check active jobs first to decide job state.
It would be better if we improved it later to determine the state with the submitted job ID. 

#231

### DIFF
--- a/controllers/flinkcluster_submit_job_script.go
+++ b/controllers/flinkcluster_submit_job_script.go
@@ -72,7 +72,6 @@ function wait_for_job() {
 		# TODO: It needs to be improved to determine the job state with the submitted job id.
 		if list_jobs | grep -e "(SCHEDULED)" -e "(RUNNING)"; then
 			echo -e "\nFound an active job."
-			sleep 30
 		else
 			if list_jobs | grep "(FINISHED)"; then
 				echo -e "\nJob has completed successfully, exiting 0"
@@ -89,6 +88,7 @@ function wait_for_job() {
 			echo -e "\nUnknown job state, exiting 3"
 			return 3
 		fi
+		sleep 30
 	done
 }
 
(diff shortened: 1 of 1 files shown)


controllers/flinkcluster_submit_job_script.go
```diff
@@ -66,19 +66,23 @@ function wait_for_job() {
...
-			return 0
-		fi
-		if list_jobs | grep "(FAILED)"; then
-			echo -e "\nJob failed, exiting 1"
-			return 1
-		fi
-		if list_jobs | grep "(CANCELED)"; then
-			echo -e "\nJob has been cancelled, exiting 2"
-			return 2
+		if list_jobs | grep -e "(SCHEDULED)" -e "(RUNNING)"; then
```
functicons, line 69: Could you add a comment here for why this is necessary? Otherwise, others might change it by mistake in the future.

elanv: I have added comments.

controllers/flinkcluster_submit_job_script.go
```diff
@@ -66,19 +66,29 @@ function wait_for_job() {
...
-		if list_jobs | grep "(CANCELED)"; then
-			echo -e "\nJob has been cancelled, exiting 2"
-			return 2
+
+		# Find active job first.
+		# If the current job is restarted by the operator, there will be records of past stopped jobs.
+		# TODO: It needs to be improved to determine the job state with the submitted job id.
+		if list_jobs | grep -e "(SCHEDULED)" -e "(RUNNING)"; then
+			echo -e "\nFound an active job."
+			sleep 30
```
functicons, line 75: I prefer putting `sleep 30` to outside of if-else, because if none of of these states are matched, we don't get into a busy wait situation.

elanv: It looks better. I've fixed.

PR #245, opened 2020-05-18
Introduce update feature

### LINKED ISSUE #65: Upgrade streaming job cluster or session cluster
Currently, to upgrade a streaming job cluster or session cluster, you must delete the existing job or cluster and then create a new CR. In this case, the order is as follows.

1. Manually cancel with savepoint using Flink API
2. Delete the current flinkcluster CR
3. Locate the saved savepoint.
4. Create a new CR specifying the manually created savepoint

If the operator can automate the upgrade, it will automate the process.

Considerations by upgrade types:

- Flink image upgrade
  - If the Flink version or java version changes.
  - In this case, tear the existing cluster and re-deploy it with a new image.
- Job only upgrade
  - Restart job by replacing jar or job resources
  - In the same way as upgrading the cluster image, you can delete the cluster and deploy a new one. 
  - Alternatively, you can leave the existing cluster as is and replace only the job and resubmit it to the existing cluster. If you want to apply task-local recovery feature, you may want the latter.  It may be provided as an option for users to select as needed.

elanv:
This PR enables Flink job update by spec.job change.

- Changes 
  - Manage CR revision history with ControllerRevision resource
    - Referenced and ported kubernetes controller history management (https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/history/controller_history.go)
    - When CR is edited, new ControllerRevision will be created.
  - Automate Flink job update process
    - When new ControllerRevision is created, operator will try to update Flink job. It will take savepoint, cancel runninig jobs and replace job submitter in the update process.

- CRD changes
  - New field of FlinkClusterSpec: RevisionHistoryLimit
  - New fields of FlinkClusterStatus: CurrentRevision, UpdateRevision, ObservedGeneration, CollisionCount

Resolves #65


functicons:
Is this PR complete for a review? Asking because you marked it with "[WIP]".

elanv:
Overall, it's done.  I'm looking for further improvement but it is ready for review.

api/v1beta1/flinkcluster_types.go
```diff
@@ -560,6 +564,21 @@ type FlinkClusterStatus struct {
 	// The status of savepoint progress
 	Savepoint *SavepointStatus `json:"savepoint,omitempty"`
 
+	// CurrentRevision indicates the version of FlinkCluster.
+	CurrentRevision string `json:"currentRevision,omitempty"`
```
functicons, line 568: Add more information about this field, e.g., why it is a string, how it is generated.

api/v1beta1/flinkcluster_types.go
```diff
@@ -93,8 +93,9 @@ const (
 	SavepointStateSucceeded     = "Succeeded"
 
 	SavepointTriggerReasonUserRequested = "user requested"
-	SavepointTriggerReasonJobCancel     = "for job-cancel"
 	SavepointTriggerReasonScheduled     = "scheduled"
+	SavepointTriggerReasonJobCancel     = "for job-cancel"
```
functicons, line 97: I feel "job cancel" is better.

api/v1beta1/flinkcluster_types.go
```diff
@@ -93,8 +93,9 @@ const (
...
+	SavepointTriggerReasonJobUpdate     = "for job update"
```
functicons, line 98: "job update".

api/v1beta1/flinkcluster_types.go
```diff
@@ -560,6 +564,21 @@ type FlinkClusterStatus struct {
 	// The status of savepoint progress
 	Savepoint *SavepointStatus `json:"savepoint,omitempty"`
 
+	// CurrentRevision indicates the version of FlinkCluster.
+	CurrentRevision string `json:"currentRevision,omitempty"`
+
+	// UpdateRevision indicates the version of FlinkCluster updating.
+	UpdateRevision string `json:"updateRevision,omitempty"`
```
functicons, line 571: Maybe `NextRevision` is more clear.

api/v1beta1/flinkcluster_types.go
```diff
@@ -383,6 +384,9 @@ type FlinkClusterSpec struct {
 
 	// Config for GCP.
 	GCPConfig *GCPConfig `json:"gcpConfig,omitempty"`
+
+	// Revision history limit to keep.
```
functicons, line 388: The maximum number of revision history to keep.

elanv:
ControllerRevision stores the content of the resource which is desired to maintain history. When the resource which is desired to maintain the history is edited, operator creates new controllerRevision resource for it and updates the "updateRevision" to the name of new controllerRevision. At that time, the name of the ControllerRevision is composed with the hash string generated from the spec to store in it. Therefore the contents of the controllerRevision resources are maintained not duplicate. 

If the content of the edited resource is the same with the one of the existing ControllerRevision, operator will only update "udpateRevision" of the resource to the name of the ControllerRevision that has the same content, instead of creating new controllerRevision.

If the currentRevision and updateRevision is not equal, it is in update state. If the desired resource which rollback revision is attached as hash label is observed, then the update process is completed and the operator will change currentRevision to updateRevision.

Additionally rollback feature also can be implemented simply with ControllerRevision and update mechanism. I think rollback can be requested by attaching control annotation with the desired revision.

controllers/flinkcluster_observer.go
```diff
@@ -79,6 +83,21 @@ func (observer *ClusterStateObserver) observe(
...
+	// Revisions.
+	var observedRevisions []*appsv1.ControllerRevision
+	err = observer.observeRevisions(&observedRevisions, observedCluster)
+	if err != nil {
+		if client.IgnoreNotFound(err) != nil {
+			log.Error(err, "Failed to get the controllerRevision resource list")
+			return err
+		}
+		log.Info("Observed controllerRevision", "controllerRevision", "nil")
+		observedCluster = nil
```
functicons, line 95: Why do we set `observedCluster` to nil?

controllers/flinkcluster_updater.go
```diff
@@ -610,6 +616,20 @@ func (updater *ClusterStatusUpdater) deriveClusterStatus(
 	}
 	status.Control = controlStatus
 
+	// handle job update
```
functicons, line 619: Can we refactor this logic into a state machine style? Making states explicitly for better readability.

controllers/flinkcluster_updater.go
```diff
@@ -790,3 +816,104 @@ func getDeploymentState(deployment *appsv1.Deployment) string {
 	}
 	return v1beta1.ComponentStateNotReady
 }
+
+func (updater *ClusterStatusUpdater) syncRevisionStatus(status *v1beta1.FlinkClusterStatus) error {
```
functicons, line 820: Add a comment for this function.

controllers/flinkcluster_util.go
```diff
@@ -129,6 +140,56 @@ func getFromSavepoint(jobSpec batchv1.JobSpec) string {
 	return ""
 }
 
+func newRevision(cluster *v1beta1.FlinkCluster, revision int64, collisionCount *int32) (*appsv1.ControllerRevision, error) {
```
functicons, line 143: Add a comment for this function.

api/v1beta1/flinkcluster_types.go
elanv: Done.

api/v1beta1/flinkcluster_types.go
elanv: Done.

api/v1beta1/flinkcluster_types.go
elanv: Done.

api/v1beta1/flinkcluster_types.go
elanv: I agree. Fixed.

api/v1beta1/flinkcluster_types.go
elanv: Done.

controllers/flinkcluster_observer.go
elanv: It was incorrect code. Fixed.

controllers/flinkcluster_updater.go
elanv: Done.

controllers/flinkcluster_util.go
elanv: Done.

functicons:
@elanv  Sorry for the slow review, I need more time to digest this change. Could you add update job in the user guide? In particular, address what can be updated, what cannot be updated.

elanv:
Fine. I also think it will take more time for documents due to other work in progress. I will also include what you mentioned in the documentation.

elanv:
I improved Flink job update implementation.
Updated user doc and explained about Flink job update.
Documentation will be updated later by adding examples, etc.

controllers/flinkcluster_updater.go
elanv: Done.

docs/user_guide.md
```diff
@@ -402,3 +402,25 @@ you can see the item named "flink-pod-monitor" in the "Service Discovery" sectio
...
+To update a running Flink job's program or execution settings, you can create new savepoint, terminate the job,
+and create a new FlinkCluster with the new program, settings, and savepoint. 
+However, in some cases, you may want to update the Flink job while maintaining the logical continuity
+of the Flink job with the FlinkCluster custom resource. In this case, you can continuously update the spec.job
+of the FlinkCluster custom resource, and the Flink operator takes care of the process required to update the Flink job.
+
+There are several points to note when using the job update feature.
+* To use the job update feature, savepointsDir must be set and the value of this field cannot be deleted when updating.
+This is because the Flink operator requires the value to create a savepoint for job updates.
+* Since fromSavepoint is used only when creating FlinkCluster, it cannot be updated.
```
functicons, line 417: After updating the job, `fromSavepoint` no longer reflects the actual savepoint the job was started from. I am thinking about adding a field in the job status to reflect the actual `fromSavepoint`. With it, we can clearly distinguish that from savepoint1 to savepoint2, it was processed by job v1, from savepoint2 to savepoint3 it was processed by job v2. It might be useful for troubleshooting, say job version2 had a bug which was identified later, then we can rerun the job after fixing the bug from its savepoint2. What do you think?

docs/user_guide.md
```diff
@@ -402,3 +402,25 @@ you can see the item named "flink-pod-monitor" in the "Service Discovery" sectio
...
+* cancelRequested and savepointGeneration are not allowed to update at the same time with different fields
+due to functional characteristics.
+
+There are some behavioral characteristics in job update.
+* Whenever the spec.job is modified, the Flink operator creates a ControllerRevision resource
+that stores the changed spec. ControllerRevision can be used to check the editing history.
+* Even if the job is in the terminated state, it can be updated. Depending on the cleanUpPolicy,
```
functicons, line 424: Elaborate the different behaviors on different cleanUpPolicies.

functicons:
It would be nice if we could provide a simple command line example for job update, not everyone knows they should update the yaml and run `kubectl apply`.

docs/user_guide.md
elanv: Good idea to add status.fromSavepoint. And for rerun, do you mean to support the Flink job replay in a way that allows updating of spec.fromSavepoint?

api/v1beta1/flinkcluster_validate.go
```diff
@@ -157,12 +161,12 @@ func (v *Validator) checkCancelRequested(
...
 
-		var accept = reflect.DeepEqual(new.Spec, oldCopy.Spec)
-		if accept {
+		if reflect.DeepEqual(new.Spec, oldCopy.Spec) {
 			return true, nil
 		}
+
 		return false, fmt.Errorf(
-			"you cannot change cancelRequested with others at the same time")
+			"you cannot update cqancelRequested with others at the same time")
```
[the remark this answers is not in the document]
elanv: Done.

api/v1beta1/flinkcluster_validate.go
```diff
@@ -183,7 +197,45 @@ func (v *Validator) checkSavepointGeneration(
...
-	return reflect.DeepEqual(new.Spec, oldCopy.Spec), nil
+	if reflect.DeepEqual(new.Spec, oldCopy.Spec) {
+		return true, nil
+	}
+
+	return false, fmt.Errorf(
+		"you cannot update savepointGeneration with others at the same time")
+}
+
+// Check job update constraints.
```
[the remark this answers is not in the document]
elanv: Done.

docs/user_guide.md
```diff
@@ -402,3 +402,25 @@ you can see the item named "flink-pod-monitor" in the "Service Discovery" sectio
...
+### Flink Job update
+
+To update a running Flink job's program or execution settings, you can create new savepoint, terminate the job,
+and create a new FlinkCluster with the new program, settings, and savepoint. 
+However, in some cases, you may want to update the Flink job while maintaining the logical continuity
+of the Flink job with the FlinkCluster custom resource. In this case, you can continuously update the spec.job
+of the FlinkCluster custom resource, and the Flink operator takes care of the process required to update the Flink job.
+
+There are several points to note when using the job update feature.
+* To use the job update feature, savepointsDir must be set and the value of this field cannot be deleted when updating.
```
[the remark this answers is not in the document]
elanv: Done.

docs/user_guide.md
```diff
@@ -402,3 +402,25 @@ you can see the item named "flink-pod-monitor" in the "Service Discovery" sectio
...
+This is because the Flink operator requires the value to create a savepoint for job updates.
+* Since fromSavepoint is used only when creating FlinkCluster, it cannot be updated.
+* cancelRequested and savepointGeneration are not allowed to update at the same time with different fields
+due to functional characteristics.
+
+There are some behavioral characteristics in job update.
+* Whenever the spec.job is modified, the Flink operator creates a ControllerRevision resource
```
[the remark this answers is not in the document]
elanv: Done.

docs/user_guide.md
elanv: Done.

elanv:
> It would be nice if we could provide a simple command line example for job update, not everyone knows they should update the yaml and run `kubectl apply`.

Okay, I will update user doc more with examples.

docs/user_guide.md
functicons: Support updating `spec.fromSavepoint` would be a nice feature, but it is up to you whether you want to implement it in this PR.

docs/user_guide.md
```diff
@@ -412,15 +412,20 @@ of the Flink job with the FlinkCluster custom resource. In this case, you can co
 of the FlinkCluster custom resource, and the Flink operator takes care of the process required to update the Flink job.
 
 There are several points to note when using the job update feature.
-* To use the job update feature, savepointsDir must be set and the value of this field cannot be deleted when updating.
+* To use the job update feature, `savepointsDir` must be set and the value of this field cannot be deleted when updating.
 This is because the Flink operator requires the value to create a savepoint for job updates.
 * Since fromSavepoint is used only when creating FlinkCluster, it cannot be updated.
-* cancelRequested and savepointGeneration are not allowed to update at the same time with different fields
+* cancelRequested and savepointGeneration are not allowed to update at the same time with other fields
```
functicons, line 418: Add quote for the fields.

docs/user_guide.md
elanv: I think it is very usable feature, but currently if fromSavepont is allowed to update, new revision will be created always even if only it is updated. What do you think about moving fromSavepont to annotation?

docs/user_guide.md
elanv: Done.

docs/user_guide.md
elanv: I think it would be better to discuss this feature further and add it.

functicons: What's the problem with "new revision will be created always even if only it is updated"?

controllers/flinkcluster_submit_job_script.go
```diff
@@ -49,7 +49,7 @@ function list_jobs() {
 function check_existing_jobs() {
 	echo "Checking existing jobs..."
 	list_jobs
-	if list_jobs | grep -e "(SCHEDULED)" -e "(RUNNING)" -e "(FINISHED)" -e "(FAILED)"; then
+	if list_jobs | grep -e "(SCHEDULED)" -e "(RUNNING)"; then
```
functicons, line 52: This reminded me that we might need to review the logic of extracting Flink job ID [here](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/a8186417d92c3eab1088570baf1d23fd5e8bbfc6/controllers/flinkcluster_observer.go#L250).  We don't allow multiple jobs in RUNNING state at the same time, but when all of them are in failed/canceled/succeeded state, how do we determine which one is the current one? I think we might need to use the recorded flink job id as a reference.

api/v1beta1/flinkcluster_validate.go
```diff
@@ -183,7 +197,44 @@ func (v *Validator) checkSavepointGeneration(
...
+	if !reflect.DeepEqual(old.Spec.Job, new.Spec.Job) {
+		if old.Spec.Job.SavepointsDir == nil || *old.Spec.Job.SavepointsDir == "" {
+			return false, fmt.Errorf(
+				"updating job is not allowed when spec.job.savepointsDir was not provided")
+		}
+		oldCopy := old.DeepCopy()
+		oldCopy.Spec.Job = new.Spec.Job.DeepCopy()
+		if !reflect.DeepEqual(oldCopy.Spec, new.Spec) {
+			return false, fmt.Errorf(
+				"you cannot update fields other than spec.job, old FlinkCluster spec: %v, new FlinkCluster spec: %v", oldCopy.Spec, new.Spec)
```
functicons, line 232: Did you consider allowing update Flink image? In that case, we might need to delete the cluster then recreate. For job update, this PR reused the existing cluster if available, but if we want to support both, we might want to use the same mechanism.

docs/user_guide.md
elanv: In more detail, I'm concern about the replay use case using savepoint rather than creating new revision itself. First, I wonder if it is useful for users to record revisions when changing fromSavepoints. If job replay is for data verification or debugging, depending on the user, it may be considered unnecessary to record it as a history. Secondly, I think users may want to replay the same savepoint multiple times for data verification or debugging. In the second case, it cannot be replayed through spec update. In this case, it seems that the replay feature can be supported by annotation. For example like this.

```
annotations:
  user-control: "replay"
```

If you specify a specific savepoint:
```
annotations:
  user-control: "replay: /tmp/savepoint"
```

If it is useful to create a revision for savepoint change, I think that we may support both fromSavepoint field and replay annotation, but I'm not sure if that's okay because it's a rather duplacate feature.

api/v1beta1/flinkcluster_validate.go
elanv: I hope the Flink cluster update is also supported. However, I think it will be a little difficult to include in this PR because it seems to have to consider not only the Flink image but also the update of other fields in the JM / TM spec. I hope the overall cluster update feature is also supported.

functicons: Right, if we support updating JM / TM spec, that would be more complete, e.g., many users are interested in scaling. If asking you to consider is that, maybe we should choose to always delete and recreate cluster when implementing job update in this PR. In that way, we could reuse the same mechanism to implement updating the whole cluster. Otherwise, we would end up with 2 mechanisms. What do you think?

docs/user_guide.md
functicons: To be honest, I don't think creating a new revision for fromSavepoint is a problem. If users really want replaying multiple times, they can create multiple CRs. I feel keeping fromSavepoint in spec is better, making it an annotation only complicates things.

api/v1beta1/flinkcluster_validate.go
elanv: I think it is basically right to create a new container as a container operation method.  However, in the case of flink cluster, it seems to be a disadvantage that it is difficult to use the local recovery function that will shorten the job resume time as well as the cluster bootstrap time.  What about providing a cluster restart option like restartOnJobUpdate?

functicons: Option sounds good to me. My point is that we need to take image / JM / tm into consideration at this point.

docs/user_guide.md
elanv: Sounds good. Let's proceed like that.

api/v1beta1/flinkcluster_validate.go
elanv: Ok, I'm going to look it a little further.

docs/user_guide.md
functicons: Seems we already have `fromSavepoint` in job status. Just need to make sure it is get updated for job update.

https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/api/v1beta1/flinkcluster_types.go#L489

elanv: Enabled `fromSavepoint` updating and introduced replay feature via it.

api/v1beta1/flinkcluster_validate.go
elanv: Improved to support both option, recreating or updating cluster.

controllers/flinkcluster_submit_job_script.go
elanv: Yes, we need to improve this later if we support cluster updating by  update API.

elanv:
I have improved this PR:
* Recreate the cluster when update triggered.
* Support cluster update also.
* Job replay by updating fromSavepoint

Next, I will update the doc more.

functicons:
Thanks! Will review it later.

functicons:
A question: since you have supported updating cluster spec, but I still see validation in the code that only job spec update is allowed, did you forget to update it?

elanv:
> A question: since you have supported updating cluster spec, but I still see validation in the code that only job spec update is allowed, did you forget to update it?
 Oh I forgot it. Later will update.

elanv:
@functicons I have checked cluster spec, but it seems to me that there are no fields that require validation for update unlike job fields. I will test it for various cases.

elanv:
Will fix update validation test later additionally.

controllers/flinkcluster_reconciler.go
```diff
@@ -373,12 +420,12 @@ func (reconciler *ClusterReconciler) reconcileJob() (ctrl.Result, error) {
...
 
 	// Update or restart
-	var jobID = reconciler.getFlinkJobID()
 	if desiredJob != nil && observedJob != nil {
+		var jobID = reconciler.getFlinkJobID()
 		var restartPolicy = observed.cluster.Spec.Job.RestartPolicy
 		var observedJobStatus = observed.cluster.Status.Components.Job
 
-		if shouldRestartJob(restartPolicy, observedJobStatus) {
+		if shouldRestartJob(restartPolicy, observedJobStatus) || shouldUpdateJob(observed) {
```
functicons, line 428: If we updated both components and job, we need to wait until the cluster is up and running before restarting the job.

controllers/flinkcluster_reconciler.go
```diff
@@ -116,9 +117,16 @@ func (reconciler *ClusterReconciler) reconcileDeployment(
 	}
 
 	if desiredDeployment != nil && observedDeployment != nil {
+		if getUpdateState(reconciler.observed) == UpdateStateUpdating {
```
functicons, line 120: Add corresponding log for other update states e.g., StoppingJob.

controllers/flinkcluster_util.go
```diff
@@ -246,3 +367,87 @@ func isUserControlFinished(controlStatus *v1beta1.FlinkClusterControlStatus) boo
...
+		completedTime := tc.FromString(jobStatus.LastSavepointTime)
+		availableTime := completedTime.Add(time.Duration(int64(SavepointAgeForJobUpdateSec) * int64(time.Second)))
+		if now.Before(availableTime) {
+			return true
+		}
+	}
+	return false
+}
+
+func isComponentUpdated(component runtime.Object, cluster v1beta1.FlinkCluster) bool {
```
functicons, line 384: Add comment for this function, in particular how it measures if a component has been updated.

controllers/flinkcluster_updater.go
```diff
@@ -594,22 +618,58 @@ func (updater *ClusterStatusUpdater) deriveClusterStatus(
...
+					State:         v1beta1.SavepointStateNotTriggered,
+					TriggerReason: v1beta1.SavepointTriggerReasonJobUpdate,
+				}
+				updater.log.Info("Savepoint will be triggered for job update")
+			} else if recorded.Savepoint != nil && recorded.Savepoint.State == v1beta1.SavepointStateInProgress {
+				updater.log.Info("Savepoint is in progress")
+			} else {
+				updater.log.Info("Stopping job")
+			}
+		case UpdateStateUpdating:
```
functicons, line 662: Does this state include both components update and job update? How do you ensure the order of them?

functicons:
CI is failing, could you check the tests?

elanv:
@functicons Thanks for your review! And sorry for the delay. There are also some non-business jobs due to work from home, so I'm working lately mainly on this.  I'll check it out late today.

functicons:
No worries, take your time.

controllers/flinkcluster_converter.go
```diff
@@ -740,11 +730,27 @@ func getDesiredJob(
...
-	jobSpec *v1beta1.JobSpec, jobStatus *v1beta1.JobStatus) *string {
+func convertFromSavepoint(jobSpec *v1beta1.JobSpec, clusterStatus *v1beta1.FlinkClusterStatus) *string {
+	var jobStatus = clusterStatus.Components.Job
+	// 1) When operator should restart the  job for some reason, we want the job resume at the stopped point,
+	// so the operator will use the latest savepoint recorded in savepointLocation in job status.
+	//
+	// 2) In the case of job update
+	// 2-1) User may want to replay the job from the savepoint user provided by updating spec.fromSavepoint.
+	// 2-2) Or user may want to resume the job from the latest savepoint after job update just like when job restarting.
+	// In this case, user must make spec.fromSavepoint empty so that the operator know the job should resume from the latest savepoint.
```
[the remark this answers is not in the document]
elanv: Then we might want to set the default value of spec.job.fromSavepoint to `LATEST`, right? Otherwise, since it is optional, if the value is not set initially, if the user updates it without changing it to `LATEST`, the job will start from the beginning, so it may be different than expected.

api/v1beta1/flinkcluster_types.go
```diff
@@ -93,8 +95,9 @@ const (
 	SavepointStateSucceeded     = "Succeeded"
 
 	SavepointTriggerReasonUserRequested = "user requested"
-	SavepointTriggerReasonJobCancel     = "for job-cancel"
 	SavepointTriggerReasonScheduled     = "scheduled"
+	SavepointTriggerReasonJobCancel     = "job cancel"
+	SavepointTriggerReasonJobUpdate     = "job update"
```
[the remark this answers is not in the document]
elanv: What about "update" more generally?

api/v1beta1/flinkcluster_validate_test.go
```diff
@@ -586,6 +577,44 @@ func TestUpdateSavepointGeneration(t *testing.T) {
 	assert.Equal(t, err2, nil)
 }
 
+func TestUpdateJob(t *testing.T) {
```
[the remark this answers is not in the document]
elanv: Done.

controllers/flinkcluster_converter.go
```diff
@@ -991,6 +1001,27 @@ func convertGCPConfig(gcpConfig *v1beta1.GCPConfig) (*corev1.Volume, *corev1.Vol
 	return saVolume, saMount, saEnv
 }
 
+func getLabels(cluster v1beta1.FlinkCluster, componentLabels map[string]string) map[string]string {
```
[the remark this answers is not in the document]
elanv: Done.

controllers/flinkcluster_converter.go
```diff
@@ -105,11 +106,10 @@ func getDesiredJobManagerDeployment(
...
 	}
 	var jobManagerDeploymentName = getJobManagerDeploymentName(clusterName)
-	var labels = map[string]string{
-		"cluster":   clusterName,
-		"app":       "flink",
+	var podLabels = getLabels(*flinkCluster, map[string]string{
 		"component": "jobmanager",
-	}
+	})
+	var labels = getLabelsRevisionHashAttached(flinkCluster.Status, podLabels)
```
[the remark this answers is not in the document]
elanv: Done.

controllers/flinkcluster_converter.go
```diff
@@ -260,11 +260,10 @@ func getDesiredJobManagerService(
...
 		TargetPort: intstr.FromString("ui")}
 	var jobManagerServiceName = getJobManagerServiceName(clusterName)
-	var labels = map[string]string{
-		"cluster":   clusterName,
-		"app":       "flink",
+	var podLabels = getLabels(*flinkCluster, map[string]string{
 		"component": "jobmanager",
-	}
+	})
+	var labels = getLabelsRevisionHashAttached(flinkCluster.Status, podLabels)
```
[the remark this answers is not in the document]
elanv: Done.

controllers/flinkcluster_reconciler.go
```diff
@@ -144,6 +152,25 @@ func (reconciler *ClusterReconciler) createDeployment(
...
+	var log = reconciler.log.WithValues("component", component)
+	if isComponentUpdated(observed, *reconciler.observed.cluster) {
+		reconciler.log.Info(fmt.Sprintf("%v is already updated, no action", component))
+		return nil
+	}
+
+	var context = reconciler.context
+	var k8sClient = reconciler.k8sClient
+	log.Info("Deleting component", "component", desired)
+	err := k8sClient.Delete(context, desired)
```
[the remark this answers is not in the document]
elanv: Changed.

controllers/flinkcluster_reconciler.go
elanv: Fixed to check it by `isClusterReady`.

controllers/flinkcluster_reconciler.go
elanv: Done.

controllers/flinkcluster_util.go
elanv: Done.

controllers/flinkcluster_updater.go
elanv: Fixed also via `isClusterReady` check in reconciler.

controllers/flinkcluster_converter.go
```diff
@@ -395,11 +392,10 @@ func getDesiredTaskManagerDeployment(
...
 	}
 	var taskManagerDeploymentName = getTaskManagerDeploymentName(clusterName)
-	var labels = map[string]string{
-		"cluster":   clusterName,
-		"app":       "flink",
+	var podLabels = getLabels(*flinkCluster, map[string]string{
 		"component": "taskmanager",
-	}
+	})
+	var labels = getLabelsRevisionHashAttached(flinkCluster.Status, podLabels)
```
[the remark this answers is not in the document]
elanv: Done.

controllers/flinkcluster_converter.go
```diff
@@ -611,18 +604,15 @@ func getDesiredJob(
 	var jobManagerServiceName = clusterName + "-jobmanager"
 	var jobManagerAddress = fmt.Sprintf(
 		"%s:%d", jobManagerServiceName, *jobManagerSpec.Ports.UI)
-	var labels = map[string]string{
-		"cluster": clusterName,
-		"app":     "flink",
-	}
+	var podLabels = getLabels(*flinkCluster, nil)
+	var labels = getLabelsRevisionHashAttached(flinkCluster.Status, podLabels)
```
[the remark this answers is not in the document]
elanv: Done.

controllers/flinkcluster_converter.go
```diff
@@ -740,11 +730,27 @@ func getDesiredJob(
 	return job
 }
 
-func convertFromSavepoint(
-	jobSpec *v1beta1.JobSpec, jobStatus *v1beta1.JobStatus) *string {
+func convertFromSavepoint(jobSpec *v1beta1.JobSpec, clusterStatus *v1beta1.FlinkClusterStatus) *string {
+	var jobStatus = clusterStatus.Components.Job
+	// 1) When operator should restart the  job for some reason, we want the job resume at the stopped point,
```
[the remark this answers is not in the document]
elanv: Done.

controllers/flinkcluster_converter.go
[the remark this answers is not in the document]
elanv: Currently fixed this as your suggestion.

controllers/flinkcluster_converter.go
```diff
@@ -106,10 +106,9 @@ func getDesiredJobManagerDeployment(
 		ports = append(ports, corev1.ContainerPort{Name: port.Name, ContainerPort: port.ContainerPort, Protocol: corev1.Protocol(port.Protocol)})
 	}
 	var jobManagerDeploymentName = getJobManagerDeploymentName(clusterName)
-	var podLabels = getLabels(*flinkCluster, map[string]string{
-		"component": "jobmanager",
-	})
-	var labels = getLabelsRevisionHashAttached(flinkCluster.Status, podLabels)
+	var componentLabels = getComponentLabels(*flinkCluster, "jobmanager")
+	var podLabels = mergeLabels(componentLabels, nil)
```
functicons, line 110: Seems there is no need to introduce `componentLabels`, we can just use `podLabels`.

controllers/flinkcluster_converter.go
```diff
@@ -604,16 +603,17 @@ func getDesiredJob(
 	var jobManagerServiceName = clusterName + "-jobmanager"
 	var jobManagerAddress = fmt.Sprintf(
 		"%s:%d", jobManagerServiceName, *jobManagerSpec.Ports.UI)
-	var podLabels = getLabels(*flinkCluster, nil)
-	var labels = getLabelsRevisionHashAttached(flinkCluster.Status, podLabels)
+	var clusterLabels = getClusterLabels(*flinkCluster)
+	var podLabels = mergeLabels(clusterLabels, nil)
```
functicons, line 607: Remove `clusterLabels`?

controllers/flinkcluster_reconciler.go
```diff
@@ -152,7 +153,7 @@ func (reconciler *ClusterReconciler) createDeployment(
 	return err
 }
 
-func (reconciler *ClusterReconciler) updateComponent(desired runtime.Object, observed runtime.Object, component string) error {
+func (reconciler *ClusterReconciler) deleteComponent(desired runtime.Object, observed runtime.Object, component string) error {
```
functicons, line 156: s/deleteComponent/deleteOldComponent

controllers/flinkcluster_reconciler.go
```diff
@@ -401,11 +403,9 @@ func (reconciler *ClusterReconciler) reconcileJob() (ctrl.Result, error) {
...
 	// Create
 	if desiredJob != nil && observedJob == nil {
-		// If the observed Flink job status list is not nil (e.g., emtpy list),
-		// it means Flink REST API server is up and running. It is the source of
-		// truth of whether we can submit a job.
-		if observed.flinkJobList == nil {
-			log.Info("Waiting for Flink API server to be ready")
+		// Proceed job creation process when the cluster components are updated and Flink API server is ready
+		if !isClusterReady(observed) {
+			log.Info("Waiting for cluster components to be updated and Flink API server to be ready")
```
functicons, line 408: s/updated/ready, it is not necessarily to be "updated" here.

controllers/flinkcluster_util.go
functicons: Avoid using `o` which is not easy to be distinguished with `0`, s/o/obj?

controllers/flinkcluster_util.go
```diff
@@ -422,20 +426,42 @@ func isComponentUpdated(component runtime.Object, cluster v1beta1.FlinkCluster)
...
-			return false
-		}
 	}
-	return true
+	return areComponentsUpdated(components, *observed.cluster)
+}
+
+// isClusterReady checks whether cluster is ready to submit job.
+// It checks if cluster components is update with desired revision and Flink API server is ready.
+func isClusterReady(observed ObservedClusterState) bool {
```
functicons, line 452: s/isClusterReady/isFlinkApiReady? or isReadyForJobSubmission.

controllers/flinkcluster_converter.go
elanv: Done.

controllers/flinkcluster_converter.go
elanv: Done.

controllers/flinkcluster_reconciler.go
elanv: Done.

controllers/flinkcluster_reconciler.go
elanv: Done.

controllers/flinkcluster_util.go
elanv: Done.

controllers/flinkcluster_converter.go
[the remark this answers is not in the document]
elanv: > Then we might want to set the default value of `spec.job.fromSavepoint` to `LATEST`, right? Otherwise, since it is optional, if the value is not set initially, when the user updates FlinkCluster without changing it to `LATEST`, the job will start from the beginning. In my opinion, it seems natural to resume the job from the latest savepoint for update as the default behavior, what do you think?

@functicons What do you think about this?

functicons:
A question: how do you handle race condition? i.e. what happens if an update is in progress and the user tries to update it again.

elanv:
To simplify the update process, if all components are deployed with the nextRevision with the revision hash value, it is completed.

If you look at each state, `StoppingJob` state is separated to prepare the latest savepoint, and the purpose is to terminate the job resource similar to the next state, `Updating`.
`Updating` state is the step of terminating all component resources other than job resource.

Changing the FlinkCluster during the update process does not make a race condition. This is because once the update is triggered (currentRevison != nextRevision), the reconciler checks whether the component's revision is the nextRevision in the moment or deletes it, but does not wait for the update to complete with the previous nextRevision.

If some componet has already been deployed with old nextRevision in this state, it will be deleted again because it does not match the nextRevision in the moment. In this way, when all components are deployed with the current nextRevision, the update process is completed and job submission or component status check is reconciled with existing logic.

docs/user_guide.md
```diff
@@ -402,3 +402,37 @@ you can see the item named "flink-pod-monitor" in the "Service Discovery" sectio
 ### Manage savepoints
 
 See this [doc](./savepoints_guide.md) on how to manage savepoints with the operator.
+
+### Flink Job update
```
functicons, line 406: s/Flink Job update/Update Flink clusters and jobs?

docs/user_guide.md
```diff
@@ -402,3 +402,37 @@ you can see the item named "flink-pod-monitor" in the "Service Discovery" sectio
...
+However, in some cases, you may want to update the Flink job while maintaining the logical continuity
+of the Flink job with the FlinkCluster custom resource. In this case, you can continuously update
+the FlinkCluster custom resource, and the Flink operator takes care of the process required to update
+the Flink job and cluster.
+
+There are several points to note when using the job update feature.
+* To use the job update feature, `savepointsDir` must be set and the value of this field cannot be deleted when updating.
+This is because the Flink operator requires the value to create a savepoint for job updates.
+* You can replay Flink job from your desired savepoint by updating `fromSavepoint`.
+If you want to resume the updated job from the latest savepoint, set `fromSavepoint` to `LATEST`.
```
functicons, line 419: What is the expected behavior if there is no savepoint taken yet.

elanv: If the job is stopped and a savepoint cannot be created, taking savepint will be aborted and the job will be started without `--fromSavepoint`. This is also true when creating a new FlinkClsuter.

If the job is running and the savepoint creation fails, the operator will keep trying until the savepoint creation succeeds. In this case, there is a possibility that a savepoint will be created, so the user will have to check the status of the Flink cluster.

elanv: In the case of updating, It seems to be improved if jobStatus.fromSavepoint is avaiable.

docs/user_guide.md
elanv: Done.

docs/user_guide.md
elanv: Fixed convertFromSavepoint to resume the job from jobStatus.fromSavepoint, when there is no latest savepoint and jobStatus.fromSavepoint is available.

elanv:
> Then we might want to set the default value of `spec.job.fromSavepoint` to `LATEST`, right? Otherwise, since it is optional, if the value is not set initially, when the user updates FlinkCluster without changing it to `LATEST`, the job will start from the beginning. In my opinion, it seems natural to resume the job from the latest savepoint for update as the default behavior, what do you think?

@functicons Do you have any opinion regarding this?

functicons:
> @functicons Do you have any opinion regarding this?

I had the same concern. It is a good idea to set LATEST as the default value, but we need to document the behavior well in CRD doc.

functicons:
> > @functicons Do you have any opinion regarding this?
> 
> I had the same concern. It is a good idea to set LATEST as the default value, but we need to document the behavior well in CRD doc.

Wait, maybe in this case we might still set empty string as the default value, but the behavior is actually LATEST, we just need to document it well. Which way do you prefer?

docs/user_guide.md
functicons: We need to add this information in the user guide, so that users can better understand the expected behavior.

elanv:
> > > @functicons Do you have any opinion regarding this?
> > 
> > 
> > I had the same concern. It is a good idea to set LATEST as the default value, but we need to document the behavior well in CRD doc.
> 
> Wait, maybe in this case we might still set empty string as the default value, but the behavior is actually "use the latest savepoint if available", we just need to document it well. Which way do you prefer?

Firstly, if we regard empty fromSavepoint as `LATEST` only, we cannot updated FlinkCluster to start the job from the begining, right? In this case, to start the job from the beginning we should create new FlinkCluster. Is this also intended? That also looks good to me. I think it's okay to define the behavior like that. 

Second, in the case of B, do you mean method 2 only or both?

Case A: FlinkCluster created without `fromSavepoint`
To update FlinkCluster to resume from latest savepoint, we update it not specifying `fromSavepoint`

Case B: FlinkCluster created with desired `fromSavepoint`
To update FlinkCluster to resume from latest savepoint:
method 1) Update FlinkClsuter with `LATEST` `fromSavepoint`
method 2) Update FlinkClsuter removing or emptying `fromSavepoint`

functicons:
> > > > @functicons Do you have any opinion regarding this?
> > > 
> > > 
> > > I had the same concern. It is a good idea to set LATEST as the default value, but we need to document the behavior well in CRD doc.
> > 
> > 
> > Wait, maybe in this case we might still set empty string as the default value, but the behavior is actually "use the latest savepoint if available", we just need to document it well. Which way do you prefer?
> 
> Firstly, if we regard empty fromSavepoint as `LATEST` only, we cannot update FlinkCluster to start the job from the begining, right? In this case, to start the job from the beginning we should create new FlinkCluster. Is this also intended? That also looks good to me. I think it's okay to define the behavior like that.
> 
> Second, in the case of B, which method do you mean among the two? Or do you mean both?
> 
> Case A: FlinkCluster created without `fromSavepoint`
> To update FlinkCluster to resume from latest savepoint, we update it not specifying `fromSavepoint`
> 
> Case B: FlinkCluster created with desired `fromSavepoint`
> To update FlinkCluster to resume from latest savepoint:
> method 1) Update FlinkCluster with `LATEST` `fromSavepoint`
> method 2) Update FlinkCluster removing or emptying `fromSavepoint`

I want to keep thing simple and explicit. What about keeping `fromSavepoint` optional, and add a new boolean field `fromLatestSavepoint`, if it is true, use latest savepoint whenever possible for job restart and update. Set default value to true. With this field, users can explicitly control the intended behavior.

elanv:
> > > > > @functicons Do you have any opinion regarding this?
> > > > 
> > > > 
> > > > I had the same concern. It is a good idea to set LATEST as the default value, but we need to document the behavior well in CRD doc.
> > > 
> > > 
> > > Wait, maybe in this case we might still set empty string as the default value, but the behavior is actually "use the latest savepoint if available", we just need to document it well. Which way do you prefer?
> > 
> > 
> > Firstly, if we regard empty fromSavepoint as `LATEST` only, we cannot update FlinkCluster to start the job from the begining, right? In this case, to start the job from the beginning we should create new FlinkCluster. Is this also intended? That also looks good to me. I think it's okay to define the behavior like that.
> > Second, in the case of B, which method do you mean among the two? Or do you mean both?
> > Case A: FlinkCluster created without `fromSavepoint`
> > To update FlinkCluster to resume from latest savepoint, we update it not specifying `fromSavepoint`
> > Case B: FlinkCluster created with desired `fromSavepoint`
> > To update FlinkCluster to resume from latest savepoint:
> > method 1) Update FlinkCluster with `LATEST` `fromSavepoint`
> > method 2) Update FlinkCluster removing or emptying `fromSavepoint`
> 
> I want to keep thing simple and explicit. What about keeping `fromSavepoint` optional, and add a new boolean field `fromLatestSavepoint`, if it is true, use latest savepoint whenever possible for job restart and update. Set default value to true. With this field, users can explicitly control the intended behavior.

Moving the spec's fromSavepoint to annotation would also be a method to do it. In this case, the job always resumes from the latest savepoint when updating. Only when `replya` annotation is attached job will start over or resume.  If its value is not specified, the job will replay again from the begining, or will resume from the specified value. Which method do you prefer?

functicons:
gcbrun

api/v1beta1/flinkcluster_types.go
```diff
@@ -391,6 +394,9 @@ type FlinkClusterSpec struct {
 
 	// Config for GCP.
 	GCPConfig *GCPConfig `json:"gcpConfig,omitempty"`
+
+	// The maximum number of revision history to keep.
+	RevisionHistoryLimit *int32 `json:"revisionHistoryLimit,omitempty"`
```
functicons, line 399: What's the default value? Also need to add it to the CRD doc.

api/v1beta1/flinkcluster_types.go
```diff
@@ -568,6 +574,22 @@ type FlinkClusterStatus struct {
 	// The status of savepoint progress
 	Savepoint *SavepointStatus `json:"savepoint,omitempty"`
 
+	// When the controller creates new ControllerRevision, it generates hash string from the FlinkCluster spec
+	// which is to be stored in ControllerRevision and uses it to compose the ControllerRevision name.
+	// Then the controller updates nextRevision to the ControllerRevision name.
+	// When update process is completed, the controller updates currentRevision as nextRevision.
+
+	// CurrentRevision indicates the version of FlinkCluster.
+	CurrentRevision string `json:"currentRevision,omitempty"`
```
functicons, line 583: Need to add all the new fields in the CRD doc.

api/v1beta1/flinkcluster_types.go
```diff
@@ -568,6 +574,22 @@ type FlinkClusterStatus struct {
 	// The status of savepoint progress
 	Savepoint *SavepointStatus `json:"savepoint,omitempty"`
 
+	// When the controller creates new ControllerRevision, it generates hash string from the FlinkCluster spec
+	// which is to be stored in ControllerRevision and uses it to compose the ControllerRevision name.
+	// Then the controller updates nextRevision to the ControllerRevision name.
```
functicons, line 579: I don't see where the controller updates status.nextRevision, could you point me to it?

docs/user_guide.md
```diff
@@ -402,3 +402,40 @@ you can see the item named "flink-pod-monitor" in the "Service Discovery" sectio
...
+This is because the Flink operator requires it to create a savepoint for job updates.
+* You can resume Flink job from your desired savepoint by updating `fromSavepoint`.
+If you want to resume the updated job from the latest savepoint, `fromSavepoint` must be unspecified.
+* `cancelRequested` and `savepointGeneration` are not allowed to update at the same time with other fields
+due to functional characteristics.
+
+There are some behavioral characteristics in update.
+* Whenever the FlinkCluster spec is updated, the Flink operator creates
+a [ControllerRevision](https://godoc.org/k8s.io/api/apps/v1#ControllerRevision) resource
+that stores the changed spec. ControllerRevisions can be used to check the editing history.
```
functicons, line 426: Add command-line examples on how to check the history.

controllers/flinkcluster_util.go
```diff
@@ -129,6 +171,70 @@ func getFromSavepoint(jobSpec batchv1.JobSpec) string {
 	return ""
 }
 
+// newRevision generates FlinkClusterSpec patch and makes new child ControllerRevision resource with it.
+func newRevision(cluster *v1beta1.FlinkCluster, revision int64, collisionCount *int32) (*appsv1.ControllerRevision, error) {
```
functicons, line 175: Can we add unit tests for these util functions?

api/v1beta1/flinkcluster_types.go
elanv: https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/2ca1acecdf486c054b14993d48c598d124b95c5a/controllers/flinkcluster_updater.go#L930-L935

api/v1beta1/flinkcluster_types.go
elanv: Done.

api/v1beta1/flinkcluster_types.go
elanv: Done.

docs/user_guide.md
elanv: Done.

controllers/flinkcluster_util.go
elanv: Done.

functicons:
I tested with the example provided in this PR, it worked perfectly. But then I tested with the [Kafka example](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/docs/kafka_test_guide.md). I set the `savepointsDir` to `/tmp` in both the old and new CR, then tried to update `jobSpec.parallelism`. What I observed was, the update got stuck, the savepoint status is changing among NotTriggered, InProgress, Failed. One thing which looks wrong is that the cluster state has been updated to Running, but the currentRevision and nextRevision still don't match. I expect the cluster should still be in Updating status. Could you try to reproduce the problem?

```
Status:
  Collision Count:  0
  Components:
    Config Map:
      Name:   flinkcluster-clickcount-configmap
      State:  Ready
    Job:
      Id:     6a9aac7f6da296ba8a9e261e17c713b7
      Name:   flinkcluster-clickcount-job
      State:  Running
    Job Manager Deployment:
      Name:   flinkcluster-clickcount-jobmanager
      State:  Ready
    Job Manager Service:
      Name:   flinkcluster-clickcount-jobmanager
      State:  Ready
    Task Manager Deployment:
      Name:          flinkcluster-clickcount-taskmanager
      State:         Ready
  Current Revision:  flinkcluster-clickcount-5ddf49895f
  Last Update Time:  2020-06-11T21:16:07Z
  Next Revision:     flinkcluster-clickcount-b6cdbc76d
  Savepoint:
    Job ID:          6a9aac7f6da296ba8a9e261e17c713b7
    State:           InProgress
    Trigger ID:      63811c8a323600d1053c2cb4e2256cb8
    Trigger Reason:  update
    Trigger Time:    2020-06-11T21:16:07Z
  State:             Running
...
Events:
  Type     Reason              Age                     From           Message
  ----     ------              ----                    ----           -------
  Normal   StatusUpdate        16m                     FlinkOperator  Cluster status: Creating
  Normal   StatusUpdate        16m                     FlinkOperator  JobManager deployment status: NotReady
  Normal   StatusUpdate        16m                     FlinkOperator  ConfigMap status: Ready
  Normal   StatusUpdate        16m                     FlinkOperator  JobManager service status: Ready
  Normal   StatusUpdate        16m                     FlinkOperator  TaskManager deployment status: NotReady
  Normal   StatusUpdate        16m                     FlinkOperator  JobManager deployment status changed: NotReady -> Ready
  Normal   StatusUpdate        10m                     FlinkOperator  Job status: Updating
  Normal   StatusUpdate        10m                     FlinkOperator  JobManager deployment status changed: Ready -> Updating
  Normal   StatusUpdate        10m                     FlinkOperator  ConfigMap status changed: Ready -> Updating
  Normal   StatusUpdate        9m58s (x11 over 10m)    FlinkOperator  (combined from similar events): Job status changed: Pending -> Running
  Normal   SavepointTriggered  5m17s                   FlinkOperator  Triggered savepoint for update: triggerID f187276a16790731c7b4034df4ea2a67.
  Normal   SavepointTriggered  5m17s                   FlinkOperator  Triggered savepoint for update: triggerID b0295003dfaafd4f2f7864f98dcc0afb.
  Warning  SavepointFailed     5m16s (x3 over 5m17s)   FlinkOperator  Savepoint creation failed: Savepoint error: java.util.concurrent.CompletionException: java.util.concurrent.CompletionException:...
  Normal   SavepointTriggered  5m16s                   FlinkOperator  Triggered savepoint for update: triggerID ffcc12acc2d6f91d915bf9c9afb530a0.
  Normal   SavepointTriggered  5m16s                   FlinkOperator  Triggered savepoint for update: triggerID 14ee73c8b7bcc66e2ae309caf14ff5a4.
  Normal   SavepointTriggered  99s (x1298 over 5m14s)  FlinkOperator  (combined from similar events): Triggered savepoint for update: triggerID 24413c70d5d5b08993f1d8cefe3a7aaf.
```

docs/user_guide.md
```diff
@@ -402,3 +402,46 @@ you can see the item named "flink-pod-monitor" in the "Service Discovery" sectio
...
+### Update Flink clusters and jobs
+
+To update a running Flink job's program or execution settings, you can create new savepoint, terminate the job,
+and create a new FlinkCluster with the new program, settings, and savepoint. 
+However, in some cases, you may want to update the Flink job while maintaining the logical continuity
+of the Flink job with the FlinkCluster custom resource. In this case, you can continuously update
+the FlinkCluster custom resource, and the Flink operator takes care of the process required to update
+the Flink job and cluster.
+
+There are several points to note when using the job update feature.
```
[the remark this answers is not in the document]
elanv: I will validate that the name cannot be changed.

elanv:
@functicons As long as job is running, the operator keep trying to take savepoint and it is intended. Usually, in that case, it is still possible to create a savepoint, and the user is likely to solve the problem, such as an error in setting the savepoint. If the job becomes failed in that state, the savepoint is aborted and the update proceeds.

The component state does not change to update until a savepoint is created. Looking at the event, it seems that the Running state has been maintained since it was created. What I think is the problem is a series of savepoint trigger requests, which seems to cause the flink to fail. This seems to need to control the interval between savepoint trigger requests.

https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/8243ff9d0ad2f01fc50d56de92d98e6a427da83a/controllers/flinkcluster_updater.go#L647-L658

functicons:
Is the too many triggers problem a race condition? Before the first one updates the status with trigger ID, the next has come?

elanv:
> Is the too many triggers problem a race condition? Before the first one updates the status with trigger ID, the next has come?

When I check the log, the savepoint status changes normally. It riggers savepoint, checks  state, and changes the state to failed sequentially. Savepoint was triggered successfully, but when the state is checked Flink server responses immediately with "Checkpoint was declined (tasks not ready)". It seems like it is needed to limit the interval between savepoint triggers.

docs/user_guide.md
```diff
@@ -402,3 +402,46 @@ you can see the item named "flink-pod-monitor" in the "Service Discovery" sectio
...
+If update is triggered in terminated state, all components are re-created as well.
+* When job is to be updated, the Flink operator will restore the job from the latest savepoint available
+- `savepointLocation` or `fromSavepoint` in job status.
+If those are not available, the job is restarted from the beginning.
+
+For example, you can create [wordcount job v1.9.2](../examples/update/wordcount-1.9.2.yaml)
+and update it to [wordcount job v1.9.3](../examples/update/wordcount-1.9.3.yaml) like this.
+
+```bash
+kubectl apply -f examples/job_update/wordcount-1.9.2.yaml
```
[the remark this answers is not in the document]
elanv: Fixed.

docs/user_guide.md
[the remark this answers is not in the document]
elanv: Validated not to allow changing namespaced name.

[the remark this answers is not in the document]
functicons: Actually, it cannot be validated, because if the namespaced name changed, it becomes a Create for k8s, it won't even hit the Update validation.

api/v1beta1/flinkcluster_validate.go
```diff
@@ -96,13 +102,29 @@ func (v *Validator) ValidateUpdate(old *FlinkCluster, new *FlinkCluster) error {
...
+
+	err = v.ValidateCreate(new)
+	if err != nil {
+		return err
 	}
 
 	return nil
 }
 
+func (v *Validator) validateNamespacedName(old metav1.ObjectMeta, new metav1.ObjectMeta) error {
```
functicons, line 118: This cannot be validated, because if the name changes, it becomes Create, the update validation won't even be hit.

functicons:
I tested with the Kafka example again, this time, it worked perfectly. But why do we have duplicate information about savepoint in status? One is under Job, one is a separate Savepoint? Is it possible to merge them under Job?

```
Status:
  Collision Count:  0
  Components:
    Config Map:
      Name:   flinkcluster-clickcount-configmap
      State:  Ready
    Job:
      From Savepoint:             file:/tmp/savepoint-64347a-0e3e5ec884ba
      Id:                         64347a80ca8cd013d91000b1809e7595
      Last Savepoint Time:        2020-06-12T16:44:27Z
      Last Savepoint Trigger ID:  e4a6a68c9d8c50f0d7ee57883c1eae16
      Name:                       flinkcluster-clickcount-job
      Savepoint Generation:       1
      Savepoint Location:         file:/tmp/savepoint-64347a-0e3e5ec884ba
      State:                      Running
    Job Manager Deployment:
      Name:   flinkcluster-clickcount-jobmanager
      State:  Ready
    Job Manager Service:
      Name:   flinkcluster-clickcount-jobmanager
      State:  Ready
    Task Manager Deployment:
      Name:          flinkcluster-clickcount-taskmanager
      State:         Ready
  Current Revision:  flinkcluster-clickcount-5ddf49895f
  Last Update Time:  2020-06-12T16:44:58Z
  Next Revision:     flinkcluster-clickcount-5ddf49895f
  Savepoint:
    Job ID:          64347a80ca8cd013d91000b1809e7595
    State:           Succeeded
    Trigger ID:      e4a6a68c9d8c50f0d7ee57883c1eae16
    Trigger Reason:  update
    Trigger Time:    2020-06-12T16:44:27Z
  State:             Running
Events:
  Type    Reason              Age                  From           Message
  ----    ------              ----                 ----           -------
  Normal  StatusUpdate        4m44s                FlinkOperator  Cluster status: Creating
  Normal  StatusUpdate        4m44s                FlinkOperator  JobManager deployment status: NotReady
  Normal  StatusUpdate        4m44s                FlinkOperator  ConfigMap status: Ready
  Normal  StatusUpdate        4m44s                FlinkOperator  JobManager service status: Ready
  Normal  StatusUpdate        4m44s                FlinkOperator  TaskManager deployment status: NotReady
  Normal  StatusUpdate        4m30s                FlinkOperator  JobManager deployment status changed: NotReady -> Ready
  Normal  StatusUpdate        4m19s                FlinkOperator  Cluster status changed: Creating -> Running
  Normal  StatusUpdate        4m17s                FlinkOperator  Job status: Pending
  Normal  SavepointTriggered  38s                  FlinkOperator  Triggered savepoint for update: triggerID e4a6a68c9d8c50f0d7ee57883c1eae16.
  Normal  SavepointCreated    38s                  FlinkOperator  Successfully savepoint created
  Normal  StatusUpdate        7s (x2 over 4m19s)   FlinkOperator  TaskManager deployment status changed: NotReady -> Ready
  Normal  StatusUpdate        7s (x13 over 3m53s)  FlinkOperator  (combined from similar events): Cluster status changed: Reconciling -> Running
```

functicons:
Another phenomenon I observed is that, after updating the spec, the old job resource was gone, but the job pod was left in error state:

```
$ kpods

NAMESPACE               NAME                                                        READY   STATUS    RESTARTS   AGE
default                 flinkcluster-clickcount-job-vp9nc                           0/1     Error     0          13m

$ klogs flinkcluster-clickcount-job-vp9nc

...
No running jobs.
No scheduled jobs.

No running and scheduled jobs, exiting 3
```

Not sure why they are not cleaned up with the job resource.

functicons:
In the Kafka example, I made the update process stuck in Reconciling state by updating the number of replicas of TM:

```
Status:
  Collision Count:  0
  Components:
    Config Map:
      Name:   flinkcluster-clickcount-configmap
      State:  Ready
    Job:
      Id:                         64347a80ca8cd013d91000b1809e7595
      Last Savepoint Time:        2020-06-12T16:44:27Z
      Last Savepoint Trigger ID:  e4a6a68c9d8c50f0d7ee57883c1eae16
      Name:                       flinkcluster-clickcount-job
      Savepoint Generation:       1
      Savepoint Location:         file:/tmp/savepoint-64347a-0e3e5ec884ba
      State:                      Running
    Job Manager Deployment:
      Name:   flinkcluster-clickcount-jobmanager
      State:  Ready
    Job Manager Service:
      Name:   flinkcluster-clickcount-jobmanager
      State:  Ready
    Task Manager Deployment:
      Name:          flinkcluster-clickcount-taskmanager
      State:         NotReady
  Current Revision:  flinkcluster-clickcount-5ddf49895f
  Last Update Time:  2020-06-12T16:59:51Z
  Next Revision:     flinkcluster-clickcount-5ddf49895f
  Savepoint:
    Job ID:          64347a80ca8cd013d91000b1809e7595
    State:           Succeeded
    Trigger ID:      e4a6a68c9d8c50f0d7ee57883c1eae16
    Trigger Reason:  update
    Trigger Time:    2020-06-12T16:44:27Z
  State:             Reconciling
```

What I did was 1) in the 1st CR, set TM replicas to a high number which makes it cannot be scheduled by k8s; 2) then in the 2nd CR, update replicas to a low number. In this case, the expected behavior should be the old TM deployment is deleted and then the new TM deployment is created, but according to the log, the old one is not deleted.

Also, while the cluster state is Reconciling, the currentRevision and nextRevision have been the same.

elanv:
> What I did was 1) in the 1st CR, set TM replicas to a high number which makes it cannot be scheduled by k8s; 2) then in the 2nd CR, update replicas to a low number. In this case, the expected behavior should be the old TM deployment is deleted and then the new TM deployment is created, but according to the log, the old one is not deleted.
> 
> Also, while the cluster state is Reconciling, the currentRevision and nextRevision have been the same.

I reproduced this. It seems that we need to improve the update trigger condition. 
If an update is in progress and you try to roll back to the original spec, the revision status will look like this:

currentRevision: A --> A --> A
nextRevision: A --> B --> A

The update is triggered by comparing the spec hash value and perhaps this caused the problem. I will fix this by comparing revision number of ControllerRevision resource.

elanv:
> I tested with the Kafka example again, this time, it worked perfectly. But why do we have duplicate information about savepoint in status? One is under Job, one is a separate Savepoint? Is it possible to merge them under Job?

Currently, savepoints in progress are recorded in `savepoint` status, and completed and available savepoints are recorded in `job` status. If merged, I think that naming of savepoint related fields will be complicated, and checking the savepoint status will be somewhat confusing as the progress and completion are recorded together under job status. If you still think it is necessary to combine, is it okay to do it in another PR?

elanv:
> Another phenomenon I observed is that, after updating the spec, the old job resource was gone, but the job pod was left in error state:

> Not sure why they are not cleaned up with the job resource.

It is needed to add a delete option and fixed it.

elanv:
@functicons Thanks for the review! It seems to be a little delayed due to various things, but will improve ASAP.

functicons:
No worries, take your time! This is a big feature, so we need to review and test it thoroughly. Thanks for your contribution!

> @functicons Thanks for the review! It seems to be a little delayed due to various things, but will improve ASAP.

functicons:
> > I tested with the Kafka example again, this time, it worked perfectly. But why do we have duplicate information about savepoint in status? One is under Job, one is a separate Savepoint? Is it possible to merge them under Job?
> 
> Currently, savepoints in progress are recorded in `savepoint` status, and completed and available savepoints are recorded in `job` status. If merged, I think that naming of savepoint related fields will be complicated, and checking the savepoint status will be somewhat confusing as the progress and completion are recorded together under `job` status. If you still think it is necessary to merge them, is it okay to do it in another PR?

Sounds good. Let's think about it carefully and do make the change in another PR. Maybe merge them into `status.savepoint`?

elanv:
> Add a point that the FlinkCluster resource name must not be changed, otherwise the operator will create a new cluster instead of updating the existing one.

This message appears when updating the FlinkCluster resource name or namespace in k8s v1.15.

When updated the name from flink-1 to flink-2:
```
error: flinkclusters.flinkoperator.k8s.io "flink-2" could not be found on the server
The edits you made on deleted resources have been saved to "/var/folders/96/yp0b49hj3vbfxm0ychk_lp1h0000gn/T/kubectl-edit-z404j.yaml"
```

When updated the namespace:
```
A copy of your changes has been stored to "/var/folders/96/yp0b49hj3vbfxm0ychk_lp1h0000gn/T/kubectl-edit-b06fj.yaml"
error: the namespace from the provided object "elan2" does not match the namespace "elan". You must pass '--namespace=elan2' to perform this operation.
```

In my dev environment, the k8s api-server seems preventing update of name or namespace.
I cannot reproduce the problem. Is it because there are differences in the development environment?

elanv:
> > > I tested with the Kafka example again, this time, it worked perfectly. But why do we have duplicate information about savepoint in status? One is under Job, one is a separate Savepoint? Is it possible to merge them under Job?
> > 
> > 
> > Currently, savepoints in progress are recorded in `savepoint` status, and completed and available savepoints are recorded in `job` status. If merged, I think that naming of savepoint related fields will be complicated, and checking the savepoint status will be somewhat confusing as the progress and completion are recorded together under `job` status. If you still think it is necessary to merge them, is it okay to do it in another PR?
> 
> Sounds good. Let's think about it carefully and make the change in another PR. Maybe merge them into `status.savepoint`?

It seems to be a good way.

elanv:
Made some changes.

* Fixed checking update trigger.
* Added cluster updating state

functicons:
Thanks for the update! I will need a few days to review and test the changes.

functicons:
Just started to look at the change. Seems there is a test failure:

```
--- FAIL: TestNewRevision (0.00s)
    flinkcluster_util_test.go:159: assertion failed: 
        --- revision
        +++ expectedRevision
          v1.ControllerRevision{
                TypeMeta: v1.TypeMeta{},
                ObjectMeta: v1.ObjectMeta{
        -               Name:         "mycluster-6d95c76b9c",
        +               Name:         "mycluster-fd56cb4fb",
                        GenerateName: "",
                        Namespace:    "default",
                        ... // 5 identical fields
                        DeletionTimestamp:          nil,
                        DeletionGracePeriodSeconds: nil,
                        Labels: map[string]string{
        -                       "flinkoperator.k8s.io/hash":       "6d95c76b9c",
        +                       "flinkoperator.k8s.io/hash":       "fd56cb4fb",
                                "flinkoperator.k8s.io/managed-by": "mycluster",
                        },
```

Could you fix it? Thanks!

elanv:
@functicons Fixed test and updated the generated files after merge.

functicons:
After checking out the latest changes with

```
$ git checkout -b elanv-feature/job-update master
$ git pull https://github.com/elanv/flink-on-k8s-operator.git feature/job-update
```

`make deploy` failed:

```
$ make deploy

go: creating new go.mod: module tmp
go: found sigs.k8s.io/controller-tools/cmd/controller-gen in sigs.k8s.io/controller-tools v0.3.0
/usr/local/google/home/dagang/go/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./api/v1beta1/..." output:crd:artifacts:config=config/crd/bases
go mod tidy
kubectl apply -f config/crd/bases
The CustomResourceDefinition "flinkclusters.flinkoperator.k8s.io" is invalid: metadata.annotations: Too long: must have at most 262144 characters
make: *** [Makefile:94: install] Error 1
```

and I noticed the generated CRD yaml changed:

```
$ git status

On branch elanv-feature/job-update
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   config/crd/bases/flinkoperator.k8s.io_flinkclusters.yaml
```

Could you take a look? Thanks!

elanv:
Fixed. It seems known issue and we can use the workaround.

https://github.com/kubernetes-sigs/kubebuilder/issues/1140

Issue #252, opened 2020-05-31
Unknown job state, exiting 3


laughingman7743:
I'm running a streaming job, but sometimes the job restarts with `Unknown job state, exiting 3`. Often occurs in jobs that have been running for longer than a day.

```
Waiting for job to finish...
Waiting for response...
------------------ Running/Restarting Jobs -------------------
30.05.2020 07:11:24 : 3d40ed91bc55f4d199d2d3f44b6c549f : com.xxx.yyy.zzz.streaming.Main$ (RUNNING)
--------------------------------------------------------------
No scheduled jobs.
30.05.2020 07:11:24 : 3d40ed91bc55f4d199d2d3f44b6c549f : com.xxx.yyy.zzz.streaming.Main$ (RUNNING)

Found an active job.

Waiting for job to finish...
Waiting for response...
------------------ Running/Restarting Jobs -------------------
30.05.2020 07:11:24 : 3d40ed91bc55f4d199d2d3f44b6c549f : com.xxx.yyy.zzz.streaming.Main$ (RUNNING)
--------------------------------------------------------------
No scheduled jobs.
30.05.2020 07:11:24 : 3d40ed91bc55f4d199d2d3f44b6c549f : com.xxx.yyy.zzz.streaming.Main$ (RUNNING)

Found an active job.

Waiting for job to finish...
Waiting for response...
------------------ Running/Restarting Jobs -------------------
30.05.2020 07:11:24 : 3d40ed91bc55f4d199d2d3f44b6c549f : com.xxx.yyy.zzz.streaming.Main$ (RESTARTING)
--------------------------------------------------------------
No scheduled jobs.
30.05.2020 07:11:24 : 3d40ed91bc55f4d199d2d3f44b6c549f : com.xxx.yyy.zzz.streaming.Main$ (RUNNING)

Found an active job.

Waiting for job to finish...
Waiting for response...
------------------ Running/Restarting Jobs -------------------
30.05.2020 07:11:24 : 3d40ed91bc55f4d199d2d3f44b6c549f : com.xxx.yyy.zzz.streaming.Main$ (RESTARTING)
--------------------------------------------------------------
No scheduled jobs.

Unknown job state, exiting 3
```

Under what circumstances does this error occur? How can I find out more information about its causes?

laughingman7743:
I'm running a streaming job with the following configuration to output Savepoints to GCS.

```yaml
[...]

PR #253, opened 2020-06-01
Add restarting job status and handle it as an active job

laughingman7743:
ref: https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/issues/252

### DIFF
--- a/controllers/flinkcluster_submit_job_script.go
+++ b/controllers/flinkcluster_submit_job_script.go
@@ -70,7 +70,7 @@ function wait_for_job() {
 		# Find active job first.
 		# If the current job is restarted by the operator, there will be records of past stopped jobs.
 		# TODO: It needs to be improved to determine the job state with the submitted job id.
-		if list_jobs | grep -e "(SCHEDULED)" -e "(RUNNING)" -e "(RESTARTING)"; then
+		if list_jobs | grep -e "(SCHEDULED)" -e "(CREATED)" -e "(SUSPENDED)" -e "(RUNNING)" -e "(RESTARTING)" -e "(CANCELLING)" -e "(FAILING)" -e "(RECONCILING)"; then
 			echo -e "\nFound an active job."
 		else
 			if list_jobs | grep "(FINISHED)"; then
--- a/controllers/flinkcluster_submit_job_script.go
+++ b/controllers/flinkcluster_submit_job_script.go
@@ -49,7 +49,7 @@ function list_jobs() {
 function check_existing_jobs() {
 	echo "Checking existing jobs..."
 	list_jobs
-	if list_jobs | grep -e "(SCHEDULED)" -e "(RUNNING)" -e "(FINISHED)" -e "(FAILED)"; then
+	if list_jobs | grep -e "(SCHEDULED)" -e "(RUNNING)" -e "(RESTARTING)" -e "(FINISHED)" -e "(FAILED)"; then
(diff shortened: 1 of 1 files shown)


functicons:
Thanks for the PR. Could you sign the CLA?

laughingman7743:
Yes. I signed the CLA.

functicons:
It didn't work. Did you use different emails for git commit and GitHub? They need to be the same.

laughingman7743:
cla
It's supposed to be the same email as the github commit. 🤔

functicons:
You need this:

> Once you've signed (or fixed any issues), please reply here with @googlebot I signed it! and we'll verify it.

controllers/flinkcluster_submit_job_script.go
```diff
@@ -70,7 +70,7 @@ function wait_for_job() {
 		# Find active job first.
 		# If the current job is restarted by the operator, there will be records of past stopped jobs.
 		# TODO: It needs to be improved to determine the job state with the submitted job id.
-		if list_jobs | grep -e "(SCHEDULED)" -e "(RUNNING)"; then
+		if list_jobs | grep -e "(SCHEDULED)" -e "(RUNNING)" -e "(RESTARTING)"; then
```
functicons, line 73: Could you add check for all states defined [here](https://ci.apache.org/projects/flink/flink-docs-release-1.9/api/java/org/apache/flink/runtime/jobgraph/JobStatus.html)?  Treat `CREATED`, `CANCELLING`, `FAILING`, `RECONCILING`, `SUSPENDED` as active.

functicons: Also, for unknown job state, don't exit, it will eventually reach terminal state.

laughingman7743: Added more statuses to treat as active jobs.

laughingman7743: https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/bf30b7b88941b8491ddcaa2935bfa4c97a409784/controllers/flinkcluster_submit_job_script.go#L52
Do I need to add some status here as well?

functicons: I mean for line 89, don't exit for unknown state, it will retry later.

laughingman7743: All right. 👍

functicons: Right, you need to add them to line 52 as well.

laughingman7743: I added it there as well. I added a line break because it was long on the side.

controllers/flinkcluster_submit_job_script.go
```diff
@@ -49,7 +49,8 @@ function list_jobs() {
 function check_existing_jobs() {
 	echo "Checking existing jobs..."
 	list_jobs
-	if list_jobs | grep -e "(SCHEDULED)" -e "(RUNNING)" -e "(FINISHED)" -e "(FAILED)"; then
+	if list_jobs | grep -e "(SCHEDULED)" -e "(CREATED)" -e "(SUSPENDED)" -e "(FINISHED)" -e "(FAILED)" \
```
functicons, line 52: Need to add `CANCELED` here.

laughingman7743: done

Issue #268, opened 2020-06-17
Support user defined annotation for k8s resources


syucream:
I want to inject some custom k8s metadata to my k8s resources to do something like using DataDog Logs with such settings,
https://docs.datadoghq.com/agent/kubernetes/log/?tab=kubernetes#configuration

PR #269, opened 2020-06-17
Support podTemplateAnnotations which injects custom annotations

syucream:
To satisfy https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/issues/268 , I want supports which accepts user defined annotations in k8s [PodTemplateSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#podtemplatespec-v1-core). It should be useful to support flexible configurations.

### DIFF
--- a/api/v1beta1/flinkcluster_types.go
+++ b/api/v1beta1/flinkcluster_types.go
@@ -212,7 +212,7 @@ type JobManagerSpec struct {
 	Sidecars []corev1.Container `json:"sidecars,omitempty"`
 
 	// JobManager Deployment pod template annotations.
-	PodTemplateAnnotations map[string]string `json:"podTemplateAnnotations,omitempty"`
+	PodAnnotations map[string]string `json:"podAnnotations,omitempty"`
 }
 
 // TaskManagerPorts defines ports of TaskManager.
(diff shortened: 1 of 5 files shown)


api/v1beta1/flinkcluster_types.go
```diff
@@ -210,6 +210,9 @@ type JobManagerSpec struct {
 	// Sidecar containers running alongside with the JobManager container in the
 	// pod.
 	Sidecars []corev1.Container `json:"sidecars,omitempty"`
+
+	// JobManager Deployment pod template annotations.
+	PodTemplateAnnotations map[string]string `json:"podTemplateAnnotations,omitempty"`
```
functicons, line 215: I think we can simplify the name, s/PodTemplateAnnotations/PodAnnotations?

functicons: Or maybe just call it Annotations.

syucream: I feel `PodAnnotations` is better so I've changed. I intentionally avoid `Annotations`, I think it'll 
 have misleading it means annotations for `Deployment`, not pod template spec. Anyway how about this?

Issue #257, opened 2020-06-01
Support batch scheduling for JobManager and TaskManager


hzxuzhonghu:
In k8s, we may meet kinds of resource-limited scenarios. In this case, it's better not to run any components, because it is likely to be useless occupying cluster resources.

So I propose making use of [Volcano](https://github.com/volcano-sh/volcano) to support gang-scheduling of all necessary managers.  And currently, spark operator has already done this.

hzxuzhonghu:
@functicons

functicons:
Sounds like a good feature request. I welcome your PR if you'd like to contribute.

hzxuzhonghu:
Yeah, thanks. I will file a pr ASAP.

hzxuzhonghu:
Just to update status, i meet a blocker as volcano depends on k8s 1.16. which has a conflict with 1.18.
Need to upgrade it there.  And i filed an issue to track https://github.com/volcano-sh/volcano/issues/849

PR #271, opened 2020-06-18
Integrate volcano to support gang scheduling for JobManager and TaskManager

hzxuzhonghu:
Fix: #257


hzxuzhonghu:
@functicons

functicons:
Thanks for the contribution! Could you finish the user guide first? I'll follow it and test.

hzxuzhonghu:
OK

docs/volcano_integration.md
```diff
@@ -0,0 +1,107 @@
+# Integration with Volcano for Batch Scheduling
+
+[Volcano](https://github.com/volcano-sh/volcano) is a batch system built on Kubernetes. It provides a suite of mechanisms
+currently missing from Kubernetes that are commonly required by many classes
+of batch & elastic workloads.
+With the integration with Volcano, Flink job and task managers can be scheduled for better scheduling efficiency.
```
functicons, line 6: For average users, they might want to learn what specific problems are solved by the integration, then they can better judge whether it is a good fit for their use case.

docs/volcano_integration.md
```diff
@@ -0,0 +1,107 @@
...
+    
+    ```bash
+   kubectl create -f https://raw.githubusercontent.com/GoogleCloudPlatform/flink-on-k8s-operator/master/config/crd/bases/flinkoperator.k8s.io_flinkclusters.yaml
+   ```
+
+4. Finally operator chart can be installed by running:
+
+	```bash
+	helm repo add flink-operator-repo https://googlecloudplatform.github.io/flink-on-k8s-operator/
+	helm install --name [RELEASE_NAME] flink-operator-repo/flink-operator --set operatorImage.name=[IMAGE_NAME]
```
functicons, line 41: This command fails for helm3, `--name` is not recognized.

docs/volcano_integration.md
```diff
@@ -0,0 +1,107 @@
...
+ 
+## Install Flink Operator with Volcano enabled
+
+## Installing the Chart
+
+1. Prepare a Flink operator image. Follow the instructions [here](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/docs/developer_guide.md#build-and-push-docker-image) to build and push an image from the source code.
+
+2. Run the bash script `update_template.sh` to update the manifest files in templates from the Flink operator source repo (This step is only required if you want to install from the local chart repo).
+
+3. Register CRD - Don't manually register CRD unless helm install below fails (You can skip this step if your helm version is v3).
```
functicons, line 31: I think it is okay to require helm 3.

docs/volcano_integration.md
```diff
@@ -0,0 +1,107 @@
...
+4. Finally operator chart can be installed by running:
+
+	```bash
+	helm repo add flink-operator-repo https://googlecloudplatform.github.io/flink-on-k8s-operator/
+	helm install --name [RELEASE_NAME] flink-operator-repo/flink-operator --set operatorImage.name=[IMAGE_NAME]
+	```
+    or to install it using local repo with command:
+
+    ```bash
+    helm install --name [RELEASE_NAME] . --set operatorImage.name=[IMAGE_NAME] --set enableBatchScheduler=true
```
functicons, line 46: Add commands on how to verify Valcano is up and running, after the installation.

docs/volcano_integration.md
```diff
@@ -0,0 +1,107 @@
...
+    ```
+
+
+# Create a sample Flink session cluster
+
+Following the guide from [Create a sample Flink cluster](./user_guide.md#create-a-sample-flink-cluster)
+
+Create a sample Flink job cluster with:
```
functicons, line 54: The following command is for session cluster instead of "job cluster". Can we add examples for both?

docs/volcano_integration.md
```diff
@@ -0,0 +1,107 @@
...
+  minMember: 2
+  minResources:
+    cpu: 400m
+    memory: 2Gi
+status:
+  phase: Running
+  running: 
+```
+
+As shown above, the podgroup has two pods in running phase and the min required number is 2, that means if the cluster has no enough resources to run both the job manager and task manager, then they are not scheduled.
```
functicons, line 100: Also describe the expected behavior if the podgroup cannot be scheduled, and what are the actions the user can take.

docs/volcano_integration.md
```diff
@@ -0,0 +1,107 @@
...
+	helm install --name [RELEASE_NAME] flink-operator-repo/flink-operator --set operatorImage.name=[IMAGE_NAME]
+	```
+    or to install it using local repo with command:
+
+    ```bash
+    helm install --name [RELEASE_NAME] . --set operatorImage.name=[IMAGE_NAME] --set enableBatchScheduler=true
+    ```
+
+
+# Create a sample Flink session cluster
```
functicons, line 50: Can we add examples for both session cluster and job cluster? IIUC, the podgroup doesn't include the job pod for job clusters, we need to mention that explicitly.

docs/volcano_integration.md
```diff
@@ -0,0 +1,107 @@
...
+kubectl apply -f https://raw.githubusercontent.com/volcano-sh/volcano/master/installer/volcano-development.yaml
+```
+
+- Install with advanced settings
+
+Please refer to [Volcano Official Guide](https://volcano.sh/docs/getting-started/)
+ 
+## Install Flink Operator with Volcano enabled
+
+## Installing the Chart
```
functicons, line 25: Can we support customizing `--enable-batch-scheduler` for `make deploy` as well? You can just follow what we do for `--watch-namespace`.

functicons:
Can we enable batch scheduler at CR level instead of operator level? I think it would be more flexible and easier for deployment.

docs/volcano_integration.md
hzxuzhonghu: make sense

docs/volcano_integration.md
hzxuzhonghu: Yes, this is copied from existing guide, i will remove it. And add specification for helm version

docs/volcano_integration.md
hzxuzhonghu: yep

docs/volcano_integration.md
hzxuzhonghu: Will look into

hzxuzhonghu:
>Can we enable batch scheduler at CR level (i.e., add a field in the CRD) instead of at operator level? I think it would be more flexible and easier for deployment.

That's great. I was thinking it would be hard to extend the API.

docs/volcano_integration.md
hzxuzhonghu: There is no diff in theory

hzxuzhonghu:
As i need to extend the FlinkCluster, so https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/pull/271#discussion_r443158670 is not necessary now.

functicons:
> As i need to extend the FlinkCluster, so [#271 (comment)](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/pull/271#discussion_r443158670) is not necessary now.

Right.

controllers/batchscheduler/factory.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/googlecloudplatform/flink-operator/controllers/batchscheduler/volcano"
+)
+
+var (
+	mutex            sync.Mutex
+	once             = sync.Once{}
+	schedulerPlugins = map[string]schedulerinterface.BatchScheduler{}
+)
+
+func lazyInitialize() {
```
functicons, line 32: How long does the init usually take? Each iteration of the reconciliation loop runs sequentially, is it really necessary to have lazy init and concurrency control for this module?

controllers/batchscheduler/interface/interface.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/googlecloudplatform/flink-operator/api/v1beta1"
+	"github.com/googlecloudplatform/flink-operator/controllers/model"
+)
+
+// BatchScheduler is the general batch scheduler interface.
+type BatchScheduler interface {
+	// Name gets the name of the scheduler
+	Name() string
+	// DoBatchScheduling reconciles batch scheduling
+	DoBatchScheduling(cluster *v1beta1.FlinkCluster, desired *model.DesiredClusterState) error
```
functicons, line 26: s/DoBatchScheduling/Schedule

controllers/batchscheduler/volcano/volcano.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/googlecloudplatform/flink-operator/api/v1beta1"
+	schedulerinterface "github.com/googlecloudplatform/flink-operator/controllers/batchscheduler/interface"
+	"github.com/googlecloudplatform/flink-operator/controllers/model"
+)
+
+const (
+	schedulerName = "volcano"
+)
+
+type VolcanoBatchScheduler struct {
```
functicons, line 39: Add a comment for this class and all public methods.

controllers/batchscheduler/factory.go
hzxuzhonghu: it does not need network connection. So it should not be a matter

hzxuzhonghu:
done

config/samples/flinkoperator_v1beta1_flinkjobcluster.yaml
```diff
@@ -20,6 +20,7 @@ spec:
   image:
     name: flink:1.8.2
   jobManager:
+    accessScope: Cluster
```
functicons, line 23: Why update this file? If not necessary, I'd suggest revert it.

controllers/batchscheduler/factory.go
functicons: Then can we remove the lazy init and mutex? let's keep it simple.

controllers/batchscheduler/volcano/volcano.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// Name returns the current scheduler name.
+func (v *VolcanoBatchScheduler) Name() string {
+	return schedulerName
+}
+
+// Schedule reconciles batch scheduling
+func (v *VolcanoBatchScheduler) Schedule(cluster *v1beta1.FlinkCluster, state *model.DesiredClusterState) error {
+	res, size := getClusterResource(state)
```
functicons, line 67: I feel we should move this function out of this file to keep this class Flink agnostic, Schedule() then takes Resources as the parameter. In this way, we will have a cleaner interface.

docs/volcano_integration.md
```diff
@@ -0,0 +1,202 @@
...
+Run the following 
+```bash
+kubectl apply -f https://raw.githubusercontent.com/volcano-sh/volcano/master/installer/volcano-development.yaml
+```
+
+- Install with advanced settings
+
+Please refer to [Volcano Official Guide](https://volcano.sh/docs/getting-started/)
+
+### Verify Volcano is up and run
```
functicons, line 24: s/run/running

docs/volcano_integration.md
```diff
@@ -0,0 +1,202 @@
...
+
+```bash
+$ kubectl get all -n volcano-system
```
functicons, line 27: The output of this command it too long, maybe just briefly mention what to check in the output.

main.go
```diff
@@ -61,7 +62,7 @@ func main() {
 		"Watch custom resources in the namespace, ignore other namespaces. If empty, all namespaces will be watched.")
 	flag.Parse()
 
-	ctrl.SetLogger(zap.Logger(true))
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
```
functicons, line 65: Should we revert this?

config/samples/flinkoperator_v1beta1_flinkjobcluster.yaml
hzxuzhonghu: This is a required field, without this, the admission denied it. Though this is not related to this, i fixed it by hand

controllers/batchscheduler/volcano/volcano.go
hzxuzhonghu: To be more extensive,  i think pass the state is more generic. We are not sure what the batch scheduler really does from the higher level

main.go
hzxuzhonghu: Not releated to this pr. Just stop using deprecated method

controllers/batchscheduler/factory.go
hzxuzhonghu: Will do that

config/samples/flinkoperator_v1beta1_flinkjobcluster.yaml
hzxuzhonghu: ···
              required:
              - accessScope
···

main.go
hzxuzhonghu: reverted

hzxuzhonghu:
@functicons please take another look

config/samples/flinkoperator_v1beta1_flinkjobcluster.yaml
```diff
@@ -37,5 +38,6 @@ spec:
     className: org.apache.flink.streaming.examples.wordcount.WordCount
     args: ["--input", "./README.txt"]
     parallelism: 2
+    restartPolicy: Never
```
hzxuzhonghu, line 41: same here, it is also a required field, if not provided. The admission controller denied create ops

config/samples/flinkoperator_v1beta1_flinkjobcluster.yaml
functicons: I checked, in the CRD doc, it says optional, but in the go struct it is required. So it seems that we intended it to be optional, but somehow the implementation was not aligned. But it is weird that in my test it doesn't fail without adding it. Any idea?

config/samples/flinkoperator_v1beta1_flinkjobcluster.yaml
functicons: This field is optional, the default is "Never", see https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/8fb6191ac0f6728e1b0e89aba49b13a7217f5996/api/v1beta1/flinkcluster_default.go#L116

docs/volcano_integration.md
```diff
@@ -0,0 +1,161 @@
...
+$ kubectl get pod -n volcano-system
+NAME                                      READY   STATUS      RESTARTS   AGE
+pod/volcano-admission-75688c79bf-b8fmj    1/1     Running     0          52s
+pod/volcano-admission-init-d684j          0/1     Completed   0          53s
+pod/volcano-controllers-d87bdbd7c-q6ds6   1/1     Running     0          52s
+pod/volcano-scheduler-5476779fd9-8rslv    1/1     Running     0          52s
+
+```
+ 
+## Install Flink Operator with Volcano enabled
```
functicons, line 36: Remove "with Volcano enabled".

docs/volcano_integration.md
```diff
@@ -0,0 +1,161 @@
...
+
+Please refer to [Deploy the operator to a Kubernetes cluster](./user_guide.md#deploy-the-operator-to-a-kubernetes-cluster)
+
+# Create a sample Flink session cluster with batch scheduling enabled
```
functicons, line 40: Sample session cluster and job cluster feel duplicate, maybe just introduce job cluster in detail, then briefly mention that "you can create a sample session cluster as well with kubectl apply -f config/samples/flinkoperator_v1beta1_flinksessioncluster_volcano.yaml".

config/samples/flinkoperator_v1beta1_flinkjobcluster.yaml
hzxuzhonghu: run kubectl get crd flinkclusters.flinkoperator.k8s.io  -oyaml to check if it is required

docs/volcano_integration.md
hzxuzhonghu: ok

config/samples/flinkoperator_v1beta1_flinkjobcluster.yaml
hzxuzhonghu: whatever, i will first revet this change. Fix it in a separate pr if needed

docs/volcano_integration.md
hzxuzhonghu: ok

config/samples/flinkoperator_v1beta1_flinkjobcluster.yaml
hzxuzhonghu: kubectl apply -f config/samples/flinkoperator_v1beta1_flinkjobcluster.yaml 
error: error validating "config/samples/flinkoperator_v1beta1_flinkjobcluster.yaml": error validating data: [ValidationError(FlinkCluster.spec.job): missing required field "restartPolicy" in io.k8s.flinkoperator.v1beta1.FlinkCluster.spec.job, ValidationError(FlinkCluster.spec.jobManager): missing required field "accessScope" in io.k8s.flinkoperator.v1beta1.FlinkCluster.spec.jobManager]; if you choose to ignore these errors, turn validation off with --validate=false

hzxuzhonghu: This is validated by kubectl, i think you are creating flinkcluster with client-go directly

hzxuzhonghu: And the default value is set by admission, so kube-apiserver will not treat is as invalid

api/v1beta1/flinkcluster_types.go
```diff
@@ -387,6 +387,10 @@ type FlinkClusterSpec struct {
 	// Flink image spec for the cluster's components.
 	Image ImageSpec `json:"image"`
 
+	// BatchSchedulerName specifies the batch scheduler name for JobManager, TaskManager.
+	// If empty, no batch scheduling is enabled.
+	BatchSchedulerName string `json:"batchSchedulerName,omitempty"`
```
functicons, line 392: Change the type to `*string` to make it optional explicitly.

hzxuzhonghu: Do we want to distinguish "" and unset?

functicons: I think both means no batch scheduling.

hzxuzhonghu: So do i need to fix it?

functicons: Yes, please. You can follow the example of `fromSavepoint` which is also an optional string field.

hzxuzhonghu: thanks for the quick response, will do

api/v1beta1/flinkcluster_types.go
```diff
@@ -389,7 +389,7 @@ type FlinkClusterSpec struct {
 
 	// BatchSchedulerName specifies the batch scheduler name for JobManager, TaskManager.
 	// If empty, no batch scheduling is enabled.
-	BatchSchedulerName string `json:"batchSchedulerName,omitempty"`
+	BatchSchedulerName *string `json:"batchSchedulerName,omitempty"`
```
functicons, line 392: After you change the type, you also need to change the places for how this field is used. The test is failing, could you fix it?

hzxuzhonghu: Sorry, i missed it

controllers/flinkcluster_reconciler.go
```diff
@@ -70,8 +70,8 @@ func (reconciler *ClusterReconciler) reconcile() (ctrl.Result, error) {
 		reconciler.log.Info("The cluster update is in progress")
 	}
 	// If batch-scheduling enabled
-	if reconciler.observed.cluster.Spec.BatchSchedulerName != "" {
-		scheduler, err := batchscheduler.GetScheduler(reconciler.observed.cluster.Spec.BatchSchedulerName)
+	if reconciler.observed.cluster.Spec.BatchSchedulerName != nil {
```
functicons, line 73: What would happen if it is ""?

hzxuzhonghu: With this, the comments is outdated, now "" means invalid scheduler name.

hzxuzhonghu: addressed,

hzxuzhonghu:
Thanks @functicons

Issue #234, opened 2020-04-21
Taskmanager supported initcontainer for load dependecy jars


Mrart:
When use initcontainer download dependency jars, must locate the jars to taskmanager pods.

functicons:
We can add the support of init container for JM and TM, which can be used for many other cases, but in terms of adding jars, it is better to create a uber jar which includes the dependencies for your job.

PR #279, opened 2020-07-10
Add support for initContainers in job and task manager.

chrismoos:
This change adds support for `initContainers` in `jobManager` and `taskManager`. This allows users to initialize the job and task managers if necessary prior to starting.

Can address issue in #234.

### DIFF
--- a/controllers/flinkcluster_converter.go
+++ b/controllers/flinkcluster_converter.go
@@ -193,6 +193,7 @@ func getDesiredJobManagerDeployment(
 	containers = append(containers, jobManagerSpec.Sidecars...)
 
 	var podSpec = corev1.PodSpec{
+		InitContainers:   convertJobManagerInitContainers(&jobManagerSpec),
 		Containers:       containers,
 		Volumes:          volumes,
 		NodeSelector:     jobManagerSpec.NodeSelector,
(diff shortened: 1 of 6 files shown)


controllers/flinkcluster_converter.go
```diff
@@ -753,6 +755,52 @@ func convertFromSavepoint(jobSpec *v1beta1.JobSpec, clusterStatus *v1beta1.Flink
 	return jobSpec.FromSavepoint
 }
 
+func convertJobManagerInitContainers(jobManagerSpec *v1beta1.JobManagerSpec) []corev1.Container {
+	var initContainers = []corev1.Container{}
+	// Add jobManagerSpec level volume mounts to each init container if there is no
+	// conflict.
+	for _, initContainer := range jobManagerSpec.InitContainers {
```
functicons, line 762: Can we extract this logic into a helper method which can be shared by all 3 methods?

controllers/flinkcluster_converter.go
```diff
@@ -753,6 +755,52 @@ func convertFromSavepoint(jobSpec *v1beta1.JobSpec, clusterStatus *v1beta1.Flink
...
+		for _, jobMount := range jobManagerSpec.VolumeMounts {
+			var conflit = false
```
functicons, line 764: s/conflit/conflict

functicons:
The test is failing, could you run `make test` and fix it?

controllers/flinkcluster_converter.go
chrismoos: This is done.

controllers/flinkcluster_converter.go
chrismoos: Fixed.

chrismoos:
@functicons Updated PR with changes, tests should be passing now as well.

PR #280, opened 2020-07-11
Add Headless AccessScope to jobmanager service spec

bnu0:
# What?

Adds `Headless` as a valid enumerated option for the `spec.jobmanager.accessScope` property in the CRD, in order to generate a [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) for the jobmanager.

# Why?

This is useful when, for example, your pod network is directly routable and you'd like to use `SRV` records (or `A` round-robin) to discover the underlying pod IPs. In this kind of topology, it is possible to avoid ingress completely yet still communicate with the jobmanager from outside the Kubernetes cluster.

Thanks!


bnu0:
@googlebot  I fixed it.

PR #282, opened 2020-07-14
Support for EnvFromSource in cluster spec

jaredstehler:
This new feature is to support EnvFromSource in cluster spec.
Prerequisites：
none
Why this feature:
Currently spec allows defining env vars but not env from source, for configmaps and secrets.
Solution:
In the Spec of FlinkCluster, we introduce a new property, "envFrom", of type []corev1.EnvFromSource.


jaredstehler:
@googlebot I signed it!

PR #283, opened 2020-07-14
Support for EnvFromSource in cluster spec

jaredstehler:
This new feature is to support EnvFromSource in cluster spec.
Prerequisites：
none
Why this feature:
Currently spec allows defining env vars but not env from source, for configmaps and secrets.
Solution:
In the Spec of FlinkCluster, we introduce a new property, "envFrom", of type []corev1.EnvFromSource.

### DIFF
--- a/api/v1beta1/flinkcluster_types.go
+++ b/api/v1beta1/flinkcluster_types.go
@@ -413,6 +413,10 @@ type FlinkClusterSpec struct {
 	// containers.
 	EnvVars []corev1.EnvVar `json:"envVars,omitempty"`
 
+	// Environment variables injected from a source, shared by all JobManager,
+	// TaskManager and job containers.
+	EnvFroms []corev1.EnvFromSource `json:"envFrom,omitempty"`
+
 	// Flink properties which are appened to flink-conf.yaml.
 	FlinkProperties map[string]string `json:"flinkProperties,omitempty"`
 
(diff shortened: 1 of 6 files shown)


jaredstehler:
> Thanks for the contribution! But you need to 1) run `make manifests generate` to generate Go and YAML files and include them in the PR; 2) update the [CRD doc](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/docs/crd.md) with the new field.
> 
> Here is an example PR of adding a new field for your reference https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/pull/279/files.

Done

api/v1beta1/flinkcluster_types.go
```diff
@@ -413,6 +413,10 @@ type FlinkClusterSpec struct {
 	// containers.
 	EnvVars []corev1.EnvVar `json:"envVars,omitempty"`
 
+	// Environment variables injected from a source, shared by all JobManager,
+	// TaskManager and job containers.
+	EnvFroms []corev1.EnvFromSource `json:"envFrom,omitempty"`
```
functicons, line 418: The json/yaml field name is not consistent with the Go field name, s/envFrom/envFroms.

jaredstehler: 👍 good catch; updating

PR #287, opened 2020-07-23
Allow customization of logging configuration 

bnu0:
# Why

Currently, flink clusters launched by the operator override the logging configs in `/opt/flink/conf` via injection of a configmap (`flink-config-volume`). So it is not possible to override logging settings with a custom docker image; we need to get the operator to do it for us instead. And frankly, thats a better option as it avoids needing to build extra images just to add configuration.

Flink itself offers to show TaskManager logs in the JM web UI. However these views do not work with the default log configuration, which outputs to std streams only. This PR makes it possible to fully control the logging behavior by configuring log4j and logback.

# What

For example, to restore logging to files (java system property `log.file` as used in the entrypoint) as well as stdout, so logging appears in *both* the flink UI as well as docker/k8s logs, etc, in your flink cluster spec, you can now add:

```yaml
logConfig:
  "log4j-console.properties": |
    rootLogger.level = INFO
    rootLogger.appenderRef.file.ref = LogFile
    rootLogger.appenderRef.console.ref = LogConsole
    appender.file.name = LogFile
    appender.file.type = File
    appender.file.append = false
    appender.file.fileName = ${sys:log.file}
    appender.file.layout.type = PatternLayout
    appender.file.layout.pattern = %d{yyyy-MM-dd HH:mm:ss,SSS} %-5p %-60c %x - %m%n
    appender.console.name = LogConsole
    appender.console.type = CONSOLE
    appender.console.layout.type = PatternLayout
    appender.console.layout.pattern = %d{yyyy-MM-dd HH:mm:ss,SSS} %-5p %-60c %x - %m%n
    logger.akka.name = akka
    logger.akka.level = INFO
    logger.kafka.name= org.apache.kafka
    logger.kafka.level = INFO
    logger.hadoop.name = org.apache.hadoop
    logger.hadoop.level = INFO
    logger.zookeeper.name = org.apache.zookeeper
    logger.zookeeper.level = INFO
    logger.netty.name = org.apache.flink.shaded.akka.org.jboss.netty.channel.DefaultChannelPipeline
    logger.netty.level = OFF
  "logback-console.xml": |
    <configuration>
        <appender name="console" class="ch.qos.logback.core.ConsoleAppender">
            <encoder>
                <pattern>%d{yyyy-MM-dd HH:mm:ss.SSS} [%thread] %-5level %logger{60} %X{sourceThread} - %msg%n</pattern>
            </encoder>
        </appender>
        <appender name="file" class="ch.qos.logback.core.FileAppender">
            <file>${log.file}</file>
            <append>false</append>
            <encoder>
                <pattern>%d{yyyy-MM-dd HH:mm:ss.SSS} [%thread] %-5level %logger{60} %X{sourceThread} - %msg%n</pattern>
            </encoder>
        </appender>
        <root level="INFO">
            <appender-ref ref="console"/>
            <appender-ref ref="file"/>
        </root>
        <logger name="akka" level="INFO" />
        <logger name="org.apache.kafka" level="INFO" />
        <logger name="org.apache.hadoop" level="INFO" />
        <logger name="org.apache.zookeeper" level="INFO" />
        <logger name="org.apache.flink.shaded.akka.org.jboss.netty.channel.DefaultChannelPipeline" level="ERROR" />
    </configuration>
```


functicons:
Thanks for the PR, will review later today.

controllers/flinkcluster_converter.go
```diff
@@ -1088,9 +1087,16 @@ log4j.logger.org.apache.flink.shaded.akka.org.jboss.netty.channel.DefaultChannel
         <appender-ref ref="console"/>
     </logger>
 </configuration>`
+)
 
+// TODO: Wouldn't it be better to create a file, put it in an operator image, and read from them?.
+// Provide logging profiles
+func getLogConf(spec v1beta1.FlinkClusterSpec) map[string]string {
+	if spec.LogConfig != nil && spec.LogConfig["log4j-console.properties"] != "" && spec.LogConfig["logback-console.xml"] != "" {
```
functicons, line 1095: What if only one of the items is empty? We need to use the default just for that item, right?

docs/crd.md
```diff
@@ -88,6 +88,9 @@ FlinkCluster
             |__ secretName
             |__ keyFile
             |__ mountPath
+    |__ logConfig
+        |__ "log4j-console.properties"
```
functicons, line 92: The type of field `logConfig` is `map[string]string`, `"log4j-console.properties"` and `"logback-console.xml"` are possible keys of the map. I think we should not add keys here, instead only document them in the field description below.

controllers/flinkcluster_converter.go
bnu0: Certainly can do that, I felt like usually it is all or nothing, and only specifying a single key is indicative of a typo/mistake. But I agree your suggestion is more general/flexible so I will change it.

bnu0:
Thanks @functicons, updated per your feedback.

PR #289, opened 2020-07-27
When create a new cr,get the terminating job status as case create error

Mrart:
Add delete false for ownerreference.

### DIFF
--- a/controllers/flinkcluster_converter.go
+++ b/controllers/flinkcluster_converter.go
@@ -562,7 +562,7 @@ func getDesiredConfigMap(
 			Namespace: clusterNamespace,
 			Name:      configMapName,
 			OwnerReferences: []metav1.OwnerReference{
-				ToJobOwnerReference(flinkCluster)},
+				ToOwnerReference(flinkCluster)},
 			Labels: labels,
 		},
 		Data: map[string]string{
@@ -716,7 +716,7 @@ func getDesiredJob(
 			Namespace: clusterNamespace,
 			Name:      jobName,
 			OwnerReferences: []metav1.OwnerReference{
-				ToOwnerReference(flinkCluster)},
+				ToJobOwnerReference(flinkCluster)},
 			Labels: jobLabels,
 		},
 		Spec: batchv1.JobSpec{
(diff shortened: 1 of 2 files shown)


controllers/flinkcluster_converter.go
```diff
@@ -562,7 +562,7 @@ func getDesiredConfigMap(
 			Namespace: clusterNamespace,
 			Name:      configMapName,
 			OwnerReferences: []metav1.OwnerReference{
-				ToOwnerReference(flinkCluster)},
+				ToJobOwnerReference(flinkCluster)},
```
functicons, line 565: I don't quite understand why we call it `ToJobOwnerReference` in a `ConfigMap` resource? What is the consequence of setting `BlockOwnerDeletion=true` for it? Will I still be able to delete the FlinkCluster CR?

Mrart: commit error

functicons:
Could you explain why we want to set `BlockOwnerDeletion=true` for the job resource? Does that mean the user won't be able to delete a FlinkCluster CR? Then what is the expected way of deleting a job cluster?

Mrart:
@functicons 
   if we set BlockOwnerDeletion=true,  After the job pod delete finished, the flinkcluster cr will deleted. this will fixed the job commit with  the seam name in CR , the new CR will get the old pod status, and the job status turning to failed

functicons:
Did you test what you said in the previous comment? Because I think setting BlockOwnerDeletion=true won't help, because we are setting it for the Job resource (which won't be deleted after the job finishes), not the Pods of the Job resource. In this case, the user has to manually delete the Job resource first, then delete the FlinkCluster. Is my understanding correct?

Mrart:
> Did you test what you said in the previous comment? Because I think setting BlockOwnerDeletion=true won't help, because we are setting it for the Job resource (which won't be deleted after the job finishes), not the Pods of the Job resource. In this case, the user has to manually delete the Job resource first, then delete the FlinkCluster. Is my understanding correct?

I have use this method in our product env, It worked. use BlockOwnerDeletion=true , the cr will deleted after the job resource clean.

functicons:
> the cr will deleted after the job resource clean.

Do you mean you need to manually delete the job resource first?

Issue #312, opened 2020-08-26
How to run JobManager and TaskManager pods as the flink user?


a-roberts:
Hey everyone, I can specify a security context using the Job CR great, e.g.

```

apiVersion: batch/v1
kind: Job
metadata:
  name: sample-job
spec:
  template:
    spec:
      # Run as flink user
      securityContext:
        runAsUser: 9999
        runAsGroup: 9999
      containers:
```

how can I do this for the JobManager/TaskManager in our FlinkCluster CR please? I've tried it in this format (spec.template.spec), and just under spec (at the FlinkCluster level), and also by adding the securityContext under jobManager and taskManager under its definition in the FlinkCluster.

Thanks!

thebalu:
Hi,
I would be interested in solving this. I have a prototype solution that exposes the setting of SecurityContexts for JM, TM and Job pods through the relevant portions of the FlinkClusterSpec. 
I have some questions about the process of contributing to this project, I'd like to get in touch with the maintainers if possible.

PR #324, opened 2020-09-17
Expose SecurityContext in FlinkClusterSpec

thebalu:
Expose the PodSecurityContext for JobManager, TaskManager, and Job pods in the FlinkClusterSpec.
This lets us run them as the Flink user, as mentioned in https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/issues/312

### DIFF
--- a/docs/crd.md
+++ b/docs/crd.md
@@ -198,6 +198,8 @@ FlinkCluster
       * **podAnnotations** (optional): Pod template annotations for the JobManager deployment.
         See [more info](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) about annotations.
       * **podLabels** (optional): Pod template labels for the JobManager deployment.
+      * **securityContext** (optional): PodSecurityContext for the JobManager pod. 
+      See [more info](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod).
     * **taskManager** (required): TaskManager spec.
       * **replicas** (required): The number of TaskManager replicas.
       * **ports** (optional): Ports that TaskManager listening on.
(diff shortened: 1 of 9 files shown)


thebalu:
Hi! This is my first time contributing to this project, so please let me know if there are any additional requirements.

I am not sure for example, if this requires an API version change; if so, could you please provide some guidance on how to do it?

thebalu:
@googlebot I signed it!

functicons (changes requested):
Thanks for the PR! 

1. Could you update the [CRD doc](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/docs/crd.md) with the new fields? 

2. Update the unit test to cover cases with and without the new fields.

3. I'd prefer keeping the sample YAMLs simple. This feature is not so frequently requested, so I don't want to confuse people with them in the sample YAMLs. For users needing it, they can go to the CRD doc.

thebalu:
@functicons I've made the requested changes, please review when you have time.

docs/crd.md
```diff
@@ -40,6 +40,7 @@ FlinkCluster
         |__ sidecars
         |__ podAnnotations
         |__ podLabels
+        |__ securityContext
```
functicons, line 43: Could you add explanation for these fields below?

functicons (changes requested):
A few minor comments.

Could you also add a section in the [user guide](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/docs/user_guide.md#how-to) on the use case and how to set these fields?

thebalu:
@functicons updated docs

functicons (approved):
Thanks!

PR #326, opened 2020-09-22
Fix cert-generation, add permission to get secret

batas:
Cert generator is unable to get secret because of missing permission.


functicons:
@batas Could you sign the CLA?

batas:
CLA signed

functicons:
@batas You need to reply `@googlebot I signed it!`.

functicons:
Are you using different emails for your GitHub account and the commit? They need to be the same.

batas:
Of course, I fixed it.

helm-chart/flink-operator/templates/generate-cert.yaml
```diff
@@ -115,6 +115,33 @@ data:
...
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: flink-operator-cert-manager
```
hongyegong, line 134: Can we call this differently? We are not using cert-manager here.
Otherwise LGTM.

batas:
Fixed

hongyegong (approved):

hongyegong:
FYI, there is an ongoing [PR](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/pull/348#issue-504263325) aiming to move the name of `flinkOperatorNamespace` a level down below, so here might become `flinkOperatorNamespace.name` after the other PR is merged.

mmazek:
What's up with all the PRs failing checks? I've looked into the logs and it does not seems as something related with the PR in particular - more like CI/CD setup issue. Can someone look into that?

rogaha:
ping

PR #331, opened 2020-09-25
Fix kube 1.18

### LINKED ISSUE #311: Kubernetes V1.18 support for CRD
I'm trying to install this operator of my kubernetes cluster running v1.18.6 however I am getting the following error
```
Error: failed to create resource: CustomResourceDefinition.apiextensions.k8s.io "flinkclusters.flinkoperator.k8s.io" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[job].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property, spec.validation.openAPIV3Schema.properties[spec].properties[taskManager].properties[sidecars].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property, spec.validation.openAPIV3Schema.properties[spec].properties[taskManager].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property, spec.validation.openAPIV3Schema.properties[spec].properties[jobManager].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property, spec.validation.openAPIV3Schema.properties[spec].properties[jobManager].properties[sidecars].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property]
```

The root cause of the issue is that V1.18 added CRDs validation [#91395](https://github.com/kubernetes/kubernetes/issues/91395)

This can be easily resolved by adding the protocol to the required fields for the item properties

```yaml
                        required:
                          - containerPort
                          - protocol
                          type: object
                        type: array
                        x-kubernetes-list-map-keys:
                        - containerPort
                        - protocol
```

EnriqueL8:
Fixes #311. Very difficult task to do with `sed` or `awk`, could be done in Golang but since `yq` used Golang it's simpler.
Could add a `yq` binary to the repo so users don't have to download it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,12 @@ run: generate fmt vet
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./api/v1beta1/..." output:crd:artifacts:config=config/crd/bases
 	go mod tidy
+	yq w -i config/crd/bases/*.yaml 'spec.validation.openAPIV3Schema.properties.spec.properties.jobManager.properties.initContainers.items.properties.ports.items.required[+]' protocol
+	yq w -i config/crd/bases/*.yaml 'spec.validation.openAPIV3Schema.properties.spec.properties.jobManager.properties.sidecars.items.properties.ports.items.required[+]' protocol
+	yq w -i config/crd/bases/*.yaml 'spec.validation.openAPIV3Schema.properties.spec.properties.job.properties.initContainers.items.properties.ports.items.required[+]' protocol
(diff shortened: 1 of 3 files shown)


functicons:
Thanks for the PR! But I feel this might not be the best solution to the problem. Can we first investigate whether we can solve the problem by upgrading controller-gen version? or using kustomize patch? Thanks!

EnriqueL8:
@functicons should I raise another PR with kustomize?

PR #334, opened 2020-10-02
support service account name

yanghui16355:
support passing service account name to operator definition

### DIFF
--- a/api/v1beta1/flinkcluster_types.go
+++ b/api/v1beta1/flinkcluster_types.go
@@ -411,9 +411,8 @@ type FlinkClusterSpec struct {
 	// Flink image spec for the cluster's components.
 	Image ImageSpec `json:"image"`
 
-	// Specifies serviceAccount name assigned to pods
-	// if empty, will use default one in namespace
-	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+	// The service account assigned to JobManager, TaskManager and Job submitter Pods. If empty, the default service account in the namespace will be used.
+	ServiceAccountName *string `json:"serviceAccountName,omitempty"`
 
 	// BatchSchedulerName specifies the batch scheduler name for JobManager, TaskManager.
 	// If empty, no batch scheduling is enabled.
--- a/api/v1beta1/flinkcluster_types.go
+++ b/api/v1beta1/flinkcluster_types.go
@@ -411,6 +411,10 @@ type FlinkClusterSpec struct {
 	// Flink image spec for the cluster's components.
 	Image ImageSpec `json:"image"`
 
+	// Specifies serviceAccount name assigned to pods
+	// if empty, will use default one in namespace
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+
 	// BatchSchedulerName specifies the batch scheduler name for JobManager, TaskManager.
 	// If empty, no batch scheduling is enabled.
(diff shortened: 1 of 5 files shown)


api/v1beta1/flinkcluster_types.go
```diff
@@ -411,6 +411,10 @@ type FlinkClusterSpec struct {
 	// Flink image spec for the cluster's components.
 	Image ImageSpec `json:"image"`
 
+	// Specifies serviceAccount name assigned to pods
```
functicons, line 414: "The service account assigned to JobManager, TaskManager and Job submitter Pods. If empty, the default service account in the namespace will be used".

api/v1beta1/flinkcluster_types.go
```diff
@@ -411,6 +411,10 @@ type FlinkClusterSpec struct {
...
+	// if empty, will use default one in namespace
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
```
functicons, line 416: Use `*string` as the type to make it optional.

controllers/flinkcluster_converter.go
```diff
@@ -87,6 +87,7 @@ func getDesiredJobManagerDeployment(
 	var clusterName = flinkCluster.ObjectMeta.Name
 	var clusterSpec = flinkCluster.Spec
 	var imageSpec = clusterSpec.Image
+	var serviceAccountSpec = clusterSpec.ServiceAccountName
```
functicons, line 90: s/serviceAccountSpec/serviceAccount

docs/crd.md
```diff
@@ -151,6 +152,8 @@ FlinkCluster
       * **pullSecrets** (optional): Secrets for image pull.
     * **batchSchedulerName** (optional): BatchSchedulerName specifies the batch scheduler name for JobManager, TaskManager.
       If empty, no batch scheduling is enabled.
+    * **serviceAccountName** (optional): serviceAccountName specifies the service account name(which must already exist in the namespace) for JobManager, TaskManager.
```
functicons, line 155: the name of the service account (which must...). If empty, the default service account in the namespace will be used.

functicons (changes requested):
Thanks! I left a few comments. Could you also add a test for the new field in https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/controllers/flinkcluster_converter_test.go?

api/v1beta1/flinkcluster_types.go
yanghui16355: fixed

api/v1beta1/flinkcluster_types.go
yanghui16355: fixed

controllers/flinkcluster_converter.go
yanghui16355: fixed

docs/crd.md
yanghui16355: fixed

yanghui16355:
unit test for service account name added

yanghui16355:
@functicons looks like all checks have passed, can you approve my PR then, thanks

functicons (approved):

yanghui16355:
@functicons one more question, how to upgrade the flink-operator with the new version(master)? thanks

PR #351, opened 2020-10-19
fix image name

yaron-idan:
It seems I've let a bug slip buy in previous PR (#348) in it broke the chart. This PR is a minor fix for that.


functicons (approved):
Thanks!

PR #354, opened 2020-10-20
Use StatefulSet instead of Deployment

### LINKED ISSUE #353: Use StatefulSet instead of Deployment for TMs and JM
Hey @functicons , When using RocksDB as state backend an SSD is optimal for performance.
When I tried to map a suitable StorageClass volume in the current deployment I failed, After an investigation, it seems there is no way to currently map a custom volume without migration from Deployment to StatefulSet.
It seems the best way to do so is to use `volumeClaimTemplates`. (https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/)
I made a branch where I changed the TMs and JMs to use StatefulSet and added `volumeClaimTemplates` to a possible configuration. 
With that change I was able to map a custom StorageAccount volume and everything seems to be working good.
StatefulSet has a parallel deployment mode that makes it deploy the pods parallel (like in Deployment) other than that I don't see a difference besides the name being more intuitive(No more random string at the end of the Pod's name)
I'll create a PR to show the needed changes. Do you see any implications of using StatefulSet rather than deployment?
The PR is: #354

shashken:
@functicons .
Closes #353


functicons:
Thanks for the PR! I need some time to digest this change. Sorry for the delayed response.

shashken:
All good @functicons tell me if there is anything you want to discuss.

functicons:
Sorry for the delayed response, I was super busy this week. I left some comments in  #353, let's first discuss it there. Thanks!

controllers/flinkcluster_observer.go
```diff
@@ -51,10 +51,10 @@ type ObservedClusterState struct {
 	cluster            *v1beta1.FlinkCluster
 	revisions          []*appsv1.ControllerRevision
 	configMap          *corev1.ConfigMap
-	jmDeployment       *appsv1.Deployment
+	jmDeployment       *appsv1.StatefulSet
```
functicons, line 54: Could you rename `Deployment` to `StatefulSet`, here and other places?

functicons (changes requested):
Could you also add the new fields in the [CRD doc](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/docs/crd.md) and add a how to section for your use case in the [user guide](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/docs/user_guide.md)?

shashken:
@functicons Changed, sorry for the delay.
There are some function names that I havn't changed like 'reconcileDeployment' do you think we should change those names as well?

functicons:
Yes, I think it is better to update all of them.

shashken:
> Yes, I think it is better to update all of them.

Done

docs/crd_v1alpha1.md
```diff
@@ -77,7 +77,7 @@ FlinkCluster
 |__ status
     |__ state
     |__ components
-        |__ jobManagerDeployment
+        |__ jobManagerStatefulSet
```
functicons, line 80: ditto.

api/v1alpha1/flinkcluster_types.go
```diff
@@ -147,6 +147,8 @@ type JobManagerSpec struct {
 	// Volume mounts in the JobManager container.
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
 
+	VolumeClaimTemplates []corev1.PersistentVolumeClaim `json:"volumeClaimTemplates,omitempty"`
```
functicons, line 150: v1alpha1 shouldn't be updated, could you revert it? Only update v1beta1.

functicons (changes requested):

api/v1alpha1/flinkcluster_types.go
shashken: @functicons fixed

shashken:
@functicons I missed the review about CRD and doc, added them now. let me know if there is something else you want me to do before merging this.

controllers/flinkcluster_controller.go
```diff
@@ -48,6 +48,8 @@ type FlinkClusterReconciler struct {
 // +kubebuilder:rbac:groups=flinkoperator.k8s.io,resources=flinkclusters/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=apps,resources=deployments/status,verbs=get
+// +kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=get;list;watch;create;update;patch;delete
```
functicons, line 51: I just tested this change with the sample cluster CRs and got permission error:

> reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.3/tools/cache/reflector.go:125: Failed to list *v1.StatefulSet: statefulsets.apps is forbidden: User "system:serviceaccount:flink-operator-system:default" cannot list resource "statefulsets" in API group "apps" at the cluster scope

In addition to this file, you need to update config/rbac/role.yaml with the required permissions for the statefulsets resource.

functicons (changes requested):

controllers/flinkcluster_controller.go
shashken: Fixed, deployed with helm do I missed it.

thanidev:
@shashken Can you please also add example sample Flink jobs and deployment involving RocksDB as state backend an SSD to config samples?

functicons:
Could you sync your repo and resolve the conflicts? @shashken

shashken:
@functicons resolved conflicts, please verify as the last PR merged today was a big one.
@thanidev Good idea, I will happily create a different PR for that, but as this one takes a bit too long I want to merge this first and then I'll create a new small PR with the example
If there aren't any problems lets merge this one, long lived PRs are not so healthy

functicons (approved):

functicons:
@shashken I have tested the PR, it worked perfectly, thanks!

burkaygur:
Looks like the current examples are not compatible with this PR. I am getting the following error:

```
2020-12-09T23:06:58.272Z	ERROR	controller-runtime.controller	Reconciler error	{"controller": "flinkcluster", "request": "default/flinksessioncluster-sample", "error": "FlinkCluster.flinkoperator.k8s.io \"flinksessioncluster-sample\" is invalid: [status.components.jobManagerStatefulSet: Required value, status.components.taskManagerStatefulSet: Required value]"}
```

thanidev:
@shashken @functicons  Can you add examples - Sample Staeful Flink jobs and deployment involving RocksDB as state backend and SSD to config samples?
More examples are better for adoption of operator.

gshen92:
@shashken @functicons Would the latest operator image v1beta1-8 on gcr.io support the changes made to use StatefulSets? Looks like there are still references to Deployment when using the sample configs for the job and session clusters.

shashken:
@gshen92 I dont know how or when those images are built, you can build the docker image from the master branch and upload to your container registry for now to see that it works :)
`make operator-image push-operator-image IMG=<your_docker_registry>/flink-google-operator:0.0.<your_version>`
And you can install the operator with the helm chart from the master branch as well.
@functicons Can probably help if you need an official image release

PR #364, opened 2020-11-04
Fix truncate history bug

DeliangFan:


functicons (changes requested):
Could you add a unit test for this method? You might consider refactoring it to make it easier to test. Thanks!

DeliangFan:
> Could you add a unit test for this method? You might consider refactoring it to make it easier to test. Thanks!

A unit test has been added, please take a review again, thanks!

functicons (approved):

bains00 (commented):
Check please

PR #371, opened 2020-11-20
[WIP] Test if serviceAccount is broken

ckdarby:
Trying to track down how serviceAccount isn't being respected


ckdarby:
@googlebot I signed it!

ckdarby:
Looks like there hasn't been a release in a while & the pipeline hasn't replaced deepcopy which was our issue.

Very annoying.

Issue #338, opened 2020-10-06
Error when creating CRD 


maver1ck:
Hi,
I'm trying to install chart version 0.2.0.
As a result I'm getting following error:
```
Error: CustomResourceDefinition.apiextensions.k8s.io "flinkclusters.flinkoperator.k8s.io" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[job].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property, spec.validation.openAPIV3Schema.properties[spec].properties[jobManager].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property, spec.validation.openAPIV3Schema.properties[spec].properties[jobManager].properties[sidecars].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property, spec.validation.openAPIV3Schema.properties[spec].properties[taskManager].properties[sidecars].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property, spec.validation.openAPIV3Schema.properties[spec].properties[taskManager].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property]
```

```
$ oc version                                                                  
Client Version: 4.5.0-0.okd-2020-10-03-012432
Server Version: 4.5.0-0.okd-2020-07-29-070316
Kubernetes Version: v1.18.3
```

functicons:
This is a known issue for Kubernetes 1.18, see #311 for more details.

Issue #266, opened 2020-06-12
make deploy failed: The CustomResourceDefinition "flinkclusters.flinkoperator.k8s.io" is invalid


abowloflrf:
Failed to deploy the operator on my minikube cluster (v1.18.2):

`make deply` error log:

```
go: creating new go.mod: module tmp
go: found sigs.k8s.io/controller-tools/cmd/controller-gen in sigs.k8s.io/controller-tools v0.3.0
/home/ruofeng/go/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./api/v1beta1/..." output:crd:artifacts:config=config/crd/bases
go mod tidy
kubectl apply -f config/crd/bases
The CustomResourceDefinition "flinkclusters.flinkoperator.k8s.io" is invalid:
* spec.validation.openAPIV3Schema.properties[spec].properties[job].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[jobManager].properties[sidecars].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[taskManager].properties[sidecars].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
make: *** [Makefile:94: install] Error 1
```

functicons:
The operator has problem with k8s 1.18, we will fix it. As a work around, you might use a lower version.

functicons:
Related info: https://github.com/SeldonIO/seldon-core/issues/1675#issuecomment-635844096

Enzo90910:
Anything I can do to help fix this? I can't really downgrade my Kubernetes on premise cluster and I would really like to give the operator a try.

bnu0:
As a temporary workaround, the crd will deploy (on 1.18 in my case) by removing the instances of `protocol` from the several `x-kubernetes-list-map-keys` arrays in the generated crd.

Enzo90910:
> As a temporary workaround, the crd will deploy (on 1.18 in my case) by removing the instances of `protocol` from the several `x-kubernetes-list-map-keys` arrays in the generated crd.

[...]

PR #381, opened 2020-12-05
Resolve CRD issue in Kubernetes v1.18

elanv:
A new version of controller-gen that solves the issue (#266 #311 #338) has been released, but it is causing another issue https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/issues/311#issuecomment-736597016. 

How about changing the controller-gen version to 0.2.4 until the CRD version is changed to v1 and the k8s.io domain problem is resolved.


elanv:
#382

functicons:
Since none of the existing versions are perfect, why do you prefer downgrading the current 0.3.0 to 0.2.4?

PR #392, opened 2021-01-11
Fix savepoint problems

shashken:
I found 2 problems related to savepoints:
1. When upgrading a job, there was no option to take a savepoint before upgrading (and using it to restore)
 _added a flag to fix this case_

2. When a cluster starts it tries to take a savepoint, the savepoint status only updates once it completes, this creates a situation where a new savepoint gets triggered while the previous one is still running, and it keeps happening if your savepoints won't finish quickly (forever)
_I solved this with another value that holds the savepoint trigger time, and an increased savepoint timeout,  so while a savepoint is still running a new one will not get triggered._

@functicons I'd love to get your feedback on this, we might want to create a stronger solution later on but for now savepoints are impossible to use with this operator if they take some time.


functicons:
Thanks for the PR, will review as soon as I get a chance.

api/v1beta1/flinkcluster_types.go
```diff
@@ -98,10 +98,11 @@ const (
 	SavepointStateFailed        = "Failed"
 	SavepointStateSucceeded     = "Succeeded"
 
-	SavepointTriggerReasonUserRequested = "user requested"
-	SavepointTriggerReasonScheduled     = "scheduled"
-	SavepointTriggerReasonJobCancel     = "job cancel"
-	SavepointTriggerReasonUpdate        = "update"
+	SavepointTriggerReasonUserRequested    = "user requested"
+	SavepointTriggerReasonScheduled        = "scheduled"
+	SavepointTriggerReasonScheduledInitial = "scheduledInitial"
```
functicons, line 103: Add a comment about this reason.

api/v1beta1/flinkcluster_types.go
```diff
@@ -347,6 +348,9 @@ type JobSpec struct {
 	// Allow non-restored state, default: false.
 	AllowNonRestoredState *bool `json:"allowNonRestoredState,omitempty"`
 
+	// Should take savepoint before upgrading the job, default: false.
+	ShouldTakeSavepointOnUpgrade *bool `json:"shouldTakeSavepointOnUpgrade,omitempty"`
```
functicons, line 352: nit: s/ShouldTakeSavepointOnUpgrade/TakeSavepointOnUpdate.

api/v1beta1/flinkcluster_types.go
```diff
@@ -567,6 +571,9 @@ type JobStatus struct {
 	// Last savepoint trigger ID.
 	LastSavepointTriggerID string `json:"lastSavepointTriggerID,omitempty"`
 
+	// Last successful or failed savepoint operation timestamp.
```
functicons, line 574: What if the operation is still in progress?

controllers/flinkcluster_reconciler.go
```diff
@@ -468,27 +468,24 @@ func (reconciler *ClusterReconciler) reconcileJob() (ctrl.Result, error) {
...
 		var restartPolicy = observed.cluster.Spec.Job.RestartPolicy
 		var recordedJobStatus = observed.cluster.Status.Components.Job
+		var jobSpec = reconciler.observed.cluster.Spec.Job
 
 		// Update or recover Flink job by restart.
-		var restartJob bool
 		if shouldUpdateJob(observed) {
 			log.Info("Job is about to be restarted to update")
-			restartJob = true
+			err := reconciler.restartJob(*jobSpec.ShouldTakeSavepointOnUpgrade)
```
functicons, line 476: I prefer keeping the existing structure and introducing a variable `takeSavepointOnUpdate` for the 2 cases.

controllers/flinkcluster_reconciler.go
```diff
@@ -468,27 +468,24 @@ func (reconciler *ClusterReconciler) reconcileJob() (ctrl.Result, error) {
...
+			err := reconciler.restartJob(false)
+			return requeueResult, err
 		}
 
 		// Trigger savepoint if required.
 		if len(jobID) > 0 {
-			if ok, savepointTriggerReason := reconciler.shouldTakeSavepoint(); ok {
+			shouldTakeSavepont, savepointTriggerReason := reconciler.shouldTakeSavepoint()
+			if shouldTakeSavepont {
+				reconciler.updateSavepointTriggerTimeStatus()
```
functicons, line 488: This method might return error.

controllers/flinkcluster_util.go
```diff
@@ -42,7 +42,7 @@ const (
 	ControlRetries            = "retries"
 	ControlMaxRetries         = "3"
 
-	SavepointTimeoutSec = 60
+	SavepointTimeoutSec = 900 // 15 mins
```
functicons, line 45: Can we make it configurable as a field in the job spec?

helm-chart/flink-operator/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 name: flink-operator
 appVersion: "1.0"
 description: A Helm chart for flink on Kubernetes operator
-version: "0.2.0"
+version: "0.2.5"
```
functicons, line 5: Why upgrade to 5?

controllers/flinkcluster_reconciler.go
```diff
@@ -744,13 +743,21 @@ func (reconciler *ClusterReconciler) shouldTakeSavepoint() (bool, string) {
 		return false, ""
 	}
 
+	var lastTriggerTime = time.Time{}
```
functicons, line 746: Extract this block into a helper method and add comments.

functicons (changes requested):
Left some comments, thanks!

api/v1beta1/flinkcluster_types.go
shashken: I'll change the comment here. This flow is still not 100% bug proof, I think more work needs to be done on savepoint flow.

controllers/flinkcluster_util.go
shashken: This might not be the best case, I increased it for the moment but I think we need to check the jobmanager's API to see SP status in the next SP PR. Do you think its bad we increased it to 15 mins? Is there a case where some1 will want to take SP every <15mins?

controllers/flinkcluster_reconciler.go
shashken: You sure? each case needs to pass a different argument to restartJob. I think its much cleaner like that

controllers/flinkcluster_util.go
functicons: This constant is actually the "minimal interval for triggering 2 savepoints", right? The name "Timeout" could be confusing, it might be mis-interpreted as "the timeout for taking a savepoint (before considering it as a failure)".

It is hard to determine the value. For example, I just took a savepoint 10 mins ago, but now I want to update my job, and I don't want to lose the state for the recent 10 mins, so I want it to take another savepoint before the update. Why do we need to introduce an arbitrary limit here?

functicons (changes requested):

controllers/flinkcluster_util.go
elanv: There are currently three time variables related to the savepoint.
* `SavepointAgeForJobUpdateSec`: savepoint age limit required for update progress
* `SavepointRequestRetryIntervalSec`: Retry interval in case of failure to create savepoint for update
* `SavepointTimeoutSec`: savepoint creation timeout

Occasionally, the savepoint creation could not proceed anymore due to an error, but the job manager returned its status normally. In that case, SavepointTimeoutSec is being used to handle the timeout.  To handle jobs that require a long time to create savepoints, it would be better to change this variable to be user-configurable and set its default value large enough.

https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/f3e73cadc44850d7abf6a6b6a612b2807afa779e/controllers/flinkcluster_util.go#L45-L51

elanv: I found that it is possible to set the checkpoint timeout with the Flink configuration. In my opinion, it would be better to remove the Flink operator's savepoint timeout handling routine to handle the second issue and guide the timeout of the Flink configuration.

note: https://ci.apache.org/projects/flink/flink-docs-release-1.12/deployment/config.html#execution-checkpointing-timeout

api/v1beta1/flinkcluster_types.go
```diff
@@ -98,10 +98,11 @@ const (
 	SavepointStateFailed        = "Failed"
 	SavepointStateSucceeded     = "Succeeded"
 
-	SavepointTriggerReasonUserRequested = "user requested"
-	SavepointTriggerReasonScheduled     = "scheduled"
-	SavepointTriggerReasonJobCancel     = "job cancel"
-	SavepointTriggerReasonUpdate        = "update"
+	SavepointTriggerReasonUserRequested    = "user requested"
+	SavepointTriggerReasonScheduled        = "scheduled"
+	SavepointTriggerReasonScheduledInitial = "scheduled initial" // The first triggered savepoint has slightly different flow
```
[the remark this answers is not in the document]
shashken: I used it to debug why I see so many triggered savepoints, but now I think it has some logic here, the first savepoint has different code flow, so it makes sense that the TriggerReason will be different. If you dont think so tell me and I'll delete it.

api/v1beta1/flinkcluster_types.go
shashken: Fixed the doc there

controllers/flinkcluster_reconciler.go
```diff
@@ -744,19 +743,31 @@ func (reconciler *ClusterReconciler) shouldTakeSavepoint() (bool, string) {
...
-	var lastTime = tc.FromString(jobStatus.LastSavepointTime)
-	var nextTime = lastTime.Add(
-		time.Duration(int64(*jobSpec.AutoSavepointSeconds) * int64(time.Second)))
+	// Scheduled, check if next trigger time arrived.
+	var nextTime = getNextOkTime(jobStatus.LastSavepointTime, int64(*jobSpec.AutoSavepointSeconds))
 	return time.Now().After(nextTime), v1beta1.SavepointTriggerReasonScheduled
 }
 
+// Convert raw time to object and add `addedSeconds` to it
+func getNextOkTime(rawTime string, addedSeconds int64) time.Time {
```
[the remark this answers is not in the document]
shashken: Changed to `getTimeAfterAddedSeconds` and added more doc

controllers/flinkcluster_util.go
shashken: I think this might be better solved in another PR, this one provides a mitigation for clusters that SP takes more than a few seconds.
I think we should discuss how we want to solve this in another issue and consider to get the info about the SP (is there an active one, was it timeout, etc..) from the jobmanager itself.
I can make this one a part of the crd for now and then later delete it when it will no longer be needed (in another PR)
WTYT @elanv @functicons

functicons: SGTM, let's address it in another PR.

functicons (approved):

controllers/flinkcluster_util.go
elanv: Sorry for the late response. When a checkpoint timeout occurs in Flink jobmanager, the savepoint state falls to "falied", so I don't think the first savepoint needs to be identified. The second issue is occurring because the default Flink checkpoint timeout is 10 minutes, but SavepointTimeoutSec is less than that. I think it's okay to handle that part in another PR.

controllers/flinkcluster_reconciler.go
```diff
@@ -468,31 +468,29 @@ func (reconciler *ClusterReconciler) reconcileJob() (ctrl.Result, error) {
...
 		// Trigger savepoint if required.
 		if len(jobID) > 0 {
-			if ok, savepointTriggerReason := reconciler.shouldTakeSavepoint(); ok {
-				newSavepointStatus, _ = reconciler.takeSavepointAsync(jobID, savepointTriggerReason)
+			shouldTakeSavepont, savepointTriggerReason := reconciler.shouldTakeSavepoint()
+			if shouldTakeSavepont {
+				err = reconciler.updateSavepointTriggerTimeStatus()
+				if err != nil {
+					newSavepointStatus, _ = reconciler.takeSavepointAsync(jobID, savepointTriggerReason)
+				}
```
elanv, line 491: @shashken @functicons 
It seems that savepoint should be triggered when `err == nil`.
When I tested I have found savepoint was not triggered and only `status.job.lastSavepointTriggerTime` was updated sometimes.

And FlinkCluster status is updated in `updateStatus` function of reconciler, therefore if you have a plan to make a new PR, it might be worth considering how to call the status update function once.

shashken: Damm I tested the version b4 the CR change with that line, sorry about that

api/v1beta1/flinkcluster_types.go
elanv: @shashken It would be nice to consider setting a default value for `takeSavepointOnUpgrade` field. Otherwise it could lead to a nil error like #408. You can easily set it with a marker like `kubebuilder:default`. 

And considering the naming consistency, the field name `TakeSavepointOnUpdate` looks better to me.

elanv: @shashken And the default value is `false` in the comment. Wouldn't it be better to set it to `true`? It seems common to restore the job from the latest savepoint.

elanv: @shashken @functicons And the default value is `false` in the comment. Wouldn't it be better to set it to `true`? It seems common to restore the job from the latest savepoint.

controllers/flinkcluster_reconciler.go
```diff
@@ -468,31 +468,29 @@ func (reconciler *ClusterReconciler) reconcileJob() (ctrl.Result, error) {
...
-		}
-		if restartJob {
-			err := reconciler.restartJob()
-			if err != nil {
-				return requeueResult, err
-			}
-			return requeueResult, nil
+			err := reconciler.restartJob(false)
+			return requeueResult, err
 		}
```
elanv, line 482: @shashken @functicons There is no need to create a savepoint here. This is because `shouldUpdateJob` checks if the latest savepoint exists. `restartJob` is just for restarting the job so there is no need to create a savepoint.

controllers/flinkcluster_reconciler.go
```diff
@@ -575,14 +573,15 @@ func (reconciler *ClusterReconciler) getFlinkJobID() string {
 	return ""
 }
 
-func (reconciler *ClusterReconciler) restartJob() error {
+func (reconciler *ClusterReconciler) restartJob(shouldTakeSavepoint bool) error {
```
elanv, line 576: @shashken @functicons `restartJob` is just for restarting the job so there is no need to trigger savepoint here.

controllers/flinkcluster_util.go
elanv: `SavepointTimeoutSec` is just savepoint timeout and `autoSavepointSeconds` is the savepoint trigger interval as  you mentioned.

PR #401, opened 2021-02-02
Fix job recovery and savepoint bug

### LINKED ISSUE #398: Job is lost after JobManager restart
Hi, 

I am running the latest version of flink-operator. I have a JobCluster, but now, if the JobManager is restarted for any reason, the FlinkCluster cannot recover. Operator shows this in the logs: 

INFO    controllers.FlinkCluster        Desired state   {"cluster": "namespace-a/job-a", "Job": "nil"}

Why is the Desired State of Job null? Also, the FlinkCluster object shows Job Status "Lost". 

In previous versions (when the submitter job never completed) the operator had no problem and would create a new Job if something went wrong. 

Thanks!

elanv:
* Job recovery bug
When job manager fails and job state falls to `Lost`, job is not properly recovered.
The job is resubmitted, but the previous job ID is not cleared, so it is considered an unexpected job and is canceled. 

* Savepoint bug
Savepoint is not triggered properly.

Resolves #398

### DIFF
--- a/controllers/flinkcluster_reconciler.go
+++ b/controllers/flinkcluster_reconciler.go
@@ -521,7 +521,7 @@ func (reconciler *ClusterReconciler) reconcileJob() (ctrl.Result, error) {
 			shouldTakeSavepont, savepointTriggerReason := reconciler.shouldTakeSavepoint()
 			if shouldTakeSavepont {
 				err = reconciler.updateSavepointTriggerTimeStatus()
-				if err != nil {
+				if err == nil {
 					newSavepointStatus, _ = reconciler.takeSavepointAsync(jobID, savepointTriggerReason)
 				}
 			}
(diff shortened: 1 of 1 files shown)


elanv:
@shashken @functicons
I have also fixed the issue https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/pull/392#discussion_r568605162 because savepoint is required for this PR.

shashken:
@elanv good job, @functicons this needs to be approved quickly, the version on master does not take SP.
I haven't done a complete test after I pushed CR changes and that line: `error != nil` broken

morelina:
I am using the commit in this PR and I am having issues when updating the job: https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/issues/408

functicons (approved):
Thanks for the fix!

pgandhijr:
I am experiencing this bug where when the job manager fails the task is submitted but it fails because it is using a a old job id.  I am using tag v1beta-9.  Is there an older stable version that is not affected by this bug that we could use?

Issue #389, opened 2020-12-27
Autoscaling enhancement


Padarn:
Hi all, 

I'm interested in autoscaling support for the operator. You can see that [ververica platform](https://www.ververica.com/blog/introducing-ververica-platform-2.2-with-autoscaling-for-apache-flink) supports this now, the mechanism I gather from their post is:

1. Monitor CPU 
2. If scale-up required, stop job with savepoint
3. Scale cluster
4. restore from savepoint

With perhaps 2 and 3 swapped to ensure smaller downtime. 

I'd be interested in adding this if it is something others would use. Any thoughts on the idea? It seems pretty straightforward to attach an HPA and use it's 'target replicas' to modify the desired number of taskmanagers.

shashken:
Hey @Padarn ,
Regarding point 2,3 I'm preparing a small PR that can tell the operator to take savepoint before cluster restart on upgrade (this is good regardless of the automatic part).
Regarding the autoscaling capability, this might be a nice idea, but this can also be a separete component (that communicates with the operator) flink scaling is a little bit complicated, and an approach that scales up a cluster based on cpu metrics alone can have no impact or even negative impact on some clusters.

Padarn:
Hey @shashken, thanks for your response. You make a good point, it can certainly be a separate component, this would be much cleaner.

Let me know when you have a PR ready, would be keen to review to get more exposure to the operator layout.

shashken:
@Padarn Done - https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/pull/392
`ShouldTakeSavepointOnUpgrade` is the flag I added. that was actually the smallest change in the PR, I fixed another savepoint issue to increase the savepoint feature stability

functicons:
Padarn@, thanks for the proposal, I would like to add auto scaling as a feature of the operator, it should be declarative for end users, a separate component would be more complex for them. It would be nice if you can contribute a PR, thanks!

Padarn:
thanks @shashken 

@functicons would be happy to work on a PR. Perhaps as there are some differing opinions on this, I will first create a POC version and ask for a review. If everyone is aligned I can clean it up to be merged.

I'll take a go at this in the weekend.

Padarn:
Hi guys. I looked into this a bit and I see there as being two options. 
[...]

PR #425, opened 2021-02-28
exposing a scale resource and way to scale jobs

Padarn:
Replaces https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/pull/394 and addresses https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/issues/389

This PR is to break down the complexity of adding autoscaling: It only attempts to add a scaling mechanism that will work for standalone Jobs. This is because the operator doesn't otherwise know what jobs are running on the cluster, or how to scale them in/out so it doesn't make much sense to try and auto scale them.

The PR adds:
- a new scale and status sub resource so that an HPA can be attached (not done in this PR)
- a new parameter inside the job spec that allows the user to specify how the parallelism should scale
- calculating parallelism from this value when job is submitted

Note some limitations:
- if the user does not set maxParalleism high enough, this won't work: unfortunately I cannot see a way to set this as an argument to Flink without explicitly doing it in code: https://ci.apache.org/projects/flink/flink-docs-stable/dev/parallel.html#setting-the-maximum-parallelism

To Do 
- [ ] documentation
- [ ] unit tests


Padarn:
@functicons I will work on the docs for this approach, but I think these are limitations we have to live with for now. Although it would be nice if it were possible to set max parallelism in the properties file.. I'll ask the flink mailing list about this.

api/v1beta1/flinkcluster_validate.go
```diff
@@ -391,7 +391,7 @@ func (v *Validator) validateTaskManager(tmSpec *TaskManagerSpec) error {
 	return nil
 }
 
-func (v *Validator) validateJob(jobSpec *JobSpec) error {
+func (v *Validator) validateJob(jobSpec *JobSpec, tmSpec *TaskManagerSpec) error {
```
functicons, line 394: Consider replacing the 2 specs with one `clusterSpec`.

api/v1beta1/flinkcluster_validate.go
```diff
@@ -406,7 +406,11 @@ func (v *Validator) validateJob(jobSpec *JobSpec) error {
 	if *jobSpec.Parallelism < 1 {
 		return fmt.Errorf("job parallelism must be >= 1")
 	}
-
+	if jobSpec.ParallelismPerTaskManager != nil    {
+		if  *jobSpec.ParallelismPerTaskManager * tmSpec.Replicas != *jobSpec.Parallelism {
```
functicons, line 410: Currently, if not specified, the default value of `jobSpec.Parallelism` is 1, but I think if we make `jobSpec.ParallelismPerTaskManager` and `jobSpec.Parallelism` mutually exclusive, it will be less confusing for users, but we need to document it well.

api/v1beta1/flinkcluster_types.go
```diff
@@ -675,6 +693,7 @@ type FlinkClusterStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.taskManager.replicas,statuspath=.status.components.taskManagerStatefulSet.replicas,selectorpath=.status.components.taskManagerStatefulSet.selector
```
functicons, line 696: I don't quite understand the meaning and purpose of this scale subresource, could you add a comment?

functicons (changes requested):

api/v1beta1/flinkcluster_validate.go
Padarn: Okay agreed on this

api/v1beta1/flinkcluster_types.go
Padarn: Adding

Padarn:
Hey @functicons I've added some unit tests (and fixes while I wrote them), and some documentation. Let me know what you think.

Padarn:
Any comments on the documentation so far? Too little, too much?

vvondruska:
Hi. Thanks a lot for this pull request. We really like the scaling functionality and we are currently testing at with a plan to use it for our work.
While testing the scaling feature we encountered a case where the operator crashed because of a nil pointer dereference - specifically here: [`controllers/flinkcluster_updater.go:392`](https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/pull/425/files#diff-ac4a4e0913a0f8fee9461f6edd793838ef0b0303c6c205b93a3389f3a88e8c7aR392).
It occurs when the operator tries to get a selector for task manager(s) of a terminating job. At that point the observed stateful set is nil and therefore its properties are no longer available.
Is this crash considered a legitimate result of providing an incorrect or incomplete configuration or should it be fixed?
Possible fixes include providing an empty label or parsing it from the recorded component.
We chose to fix it by parsing the selector from the recorded component in case it's available. I can push the fix if you'd like.

Padarn:
Hey @vvondruska thanks a lot for the feedback! Happy to have your fix, you could either push it or create a small snippet and I'll incorporate and push it. Either way works.

vvondruska:
Thank you for a quick response @Padarn. I can push the fix. I was wondering if it would be OK with you if I pushed the code directly to the repository with the scaling functionality (`Padarn:expose-scale`). I would need a temporary write privilege to do that. Or would you prefer a different way to apply the fix?

Padarn:
Makes sense to me. I sent you an invite to the repo so you can push.

vvondruska:
@googlebot I fixed it.

vvondruska:
@googlebot I consent.

Padarn:
@googlebot I consent.

mvbakker:
@Padarn @functicons is there anything we (@vvondruska and I) can do to help to get this merged in?

Padarn:
I'm happy to @mvbakker but cannot approve anything. Also happy to make any further changes if there is anything blocking.

Padarn:
Hi @functicons any change we ca nget this merged?

PR #432, opened 2021-03-06
add annotation to stateful set

yanghui16355:


yanghui16355:
This small PR is to add the annotation to the job/task managers' stateful set that it can be recognized by Spinnaker or other third-party tool which use the annotation on deployment or statefulSet.

@functicons please have a review and approve, thanks

PR #439, opened 2021-04-13
adds connection timeout to job submitter

sv3ndk:
Hi all,

While experimenting with the Flink operator, I experienced some occasional glitches when the very first `curl` done by the job submitter hangs for a very long time before retrying, instead of the usual 5 seconds. It then eventually times out, although this long pause has an unnecessary impact on the deployment duration.

```sh
Checking job manager to be ready. Will check success of 2 API calls for stable job submission.
curl -sS "http://fandom-analytics-metrics-computer-jobmanager:8081/jobs"
```

I believe this happens when the job submitter calls `curl` before the job manager is even started, in which case we're getting connection timeout (after 120s) instead of an immediate "connection refused".

After adding the `--connect-timeout 5`, all deployment continued to work successfully, and I never experienced again the hanging behaviour mentioned above.

Let me know your thoughts


sv3ndk:
@googlebot I signed it!

sv3ndk:
@hongyegong , could you please have a quick look at this minor update and let me know what you think? Thanks

PR #451, opened 2021-05-25
Fix memory divisor in calHeapSize

sharkdtu:
Currently, the calculated flink heap size is larger than it really is, As it used 1000 as divisor. 

This pr used 1024 as divisor to fix the issue.


sharkdtu:
Can one of the admins verify this patch?

PR #456, opened 2021-06-24
fix bug for ARG overload

DevineLiu:
executor  failed running [/bin/sh -c test -n "$FLINK_VERSION"] ,because  in the "FROM" stage, the arg is initialized.


DevineLiu:
An ARG declared before a FROM is outside of a build stage, so it can’t be used in any instruction after a FROM. To use the default value of an ARG declared before the first FROM use an ARG instruction without a value inside of a build stage:

DevineLiu:
document:
https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

PR #458, opened 2021-06-28
Update flink-operator.yaml for helm resource-polic

DevineLiu:
When you set the operator namespace to an existing one, helm delete will recycle the namespace, which is a dangerous operation and will cause all the resources of the namespace to be recycled.

### DIFF
--- a/helm-chart/flink-operator/templates/flink-operator.yaml
+++ b/helm-chart/flink-operator/templates/flink-operator.yaml
@@ -3,6 +3,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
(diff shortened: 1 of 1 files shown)


sv3ndk:
Hi, I'm not a maintainer of this repo, though I'm using it in production for one of my projects.
My perception is that the maintainers of this project became silent in February this year, AFAIK no issue nor PR has received any attention since then.
Where I work, we're worried about what that situation implies for us.
Our usage is too small to justify taking over the maintenance of this, although if somebody did, we would be able to continue contributing to it (I submitted already a minor PR a few months back).

PR #461, opened 2021-07-12
upgrade client-go version to 0.21.2

johnlanni:


johnlanni:
@googlebot I signed it

